PR #6, opened 2015-06-06
Add initial javascript support with jsdoc 3.

ericholscher:
This is a very first pass at adding jsdoc support.

### DIFF
--- a/autoapi/domains/javascript.py
+++ b/autoapi/domains/javascript.py
@@ -1,0 +1,201 @@
+import os
+import json
+import subprocess
+
+from sphinx.util.osutil import ensuredir
+
+from .base import AutoAPIBase, AutoAPIDomain
+from ..settings import env
+
+
+class JavaScriptDomain(AutoAPIDomain):
+
+    '''Auto API domain handler for Javascript
+
+    Parses directly from Javascript files.
+
+    :param app: Sphinx application passed in as part of the extension
+    '''
+
+    def create_class(self, data):
+        '''Return instance of class based on Javascript data
+
+        Data keys handled here:
+
+            type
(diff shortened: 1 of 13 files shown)


autoapi/domains/javascript.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def create_class(self, data):
+        '''Return instance of class based on Javascript data
+
+        Data keys handled here:
+
+            type
+                Set the object class
+
+            consts, types, vars, funcs
```
agjohnson: Looks like go boilerplate here

autoapi/domains/javascript.py
```diff
@@ -0,0 +1,201 @@
...
+
+        :param path: Path of file to read
+        '''
+        # TODO support JSON here
+        # TODO sphinx way of reporting errors in logs?
+
+        try:
+            parsed_data = json.loads(subprocess.check_output(['jsdoc', '-X', path]))
+            return parsed_data
+        except IOError:
```
agjohnson: Catch `OSError` or something from `subprocess` here, if `jsdoc` is not installed or in the PATH

autoapi/domains/javascript.py
```diff
@@ -0,0 +1,201 @@
...
+        #self.docstring = obj.get('comment', '')
+
+        self.imports = obj.get('imports', [])
+        self.children = []
+        self.parameters = map(
+            lambda n: {'name': n['name'],
+                       'type': n['type'][0]},
+            obj.get('param', [])
+        )
+
```
agjohnson: Does jsdoc give us return type/info we could add here?

PR #9, opened 2015-06-10
Plane work

ericholscher:
A bunch of changes around the refactor. Main things:

* Configurable options in the config that are passed to templates
* Tests for all current languages
* All languages passing basic tests
* Renamed "Domains" to "Mappers"
* Add working configurable templates & tests for that


autoapi/mappers/base.py
```diff
@@ -176,14 +210,19 @@ def create_class(self, obj):
...
             rst = obj.render()
             if not rst:
                 continue
 
-            detail_dir = os.path.join(root, *id.split('.'))
+            try:
+                filename = id.split('(')[0]
+            except IndexError:
+                filename = id
+            filename = filename.replace('#', '-')
```
agjohnson: What's this for? Is this .NET `Foobar#ctor` parsing? Are there more cases to be slugged here, or does it make sense to move this to the .NET mapper?

autoapi/mappers/base.py
```diff
@@ -191,6 +230,12 @@ def output_rst(self, root, source_suffix):
 
         # Render Top Index
         top_level_index = os.path.join(root, 'index.rst')
+        pages = self.objects.values()
+        # for key, item in self.objects.items():
+        #     if key.count('.') == 1:
+        #         top_level_pages.append(item)
+        #     else:
+        #         print key, key.find('.')
```
agjohnson: Commented code, can remove this.

autoapi/mappers/dotnet.py
```diff
@@ -127,63 +133,44 @@ def _recurse_ns(obj):
...
+    #         # TODO: Better way to determine suffix?
+    #         path = os.path.join(detail_dir, '%s%s' % ('index', source_suffix))
+    #         with open(path, 'w+') as detail_file:
+    #             detail_file.write(rst.encode('utf-8'))
 
+    #     # Write Indexes
+    #     top_level_index = os.path.join(root, 'index.rst')
+    #     with open(top_level_index, 'w+') as top_level_file:
+    #         content = self.jinja_env.get_template('index.rst')
+    #         top_level_file.write(content.render(pages=self.namespaces.values()))
```
agjohnson: Some more commented out code, might as well delete it

autoapi/mappers/dotnet.py
```diff
@@ -252,9 +239,10 @@ def edit_link(self):
             return '{repo}/blob/master/{path}'.format(
                 repo=repo,
                 path=path,
-                )
+            )
         except:
-            import traceback; traceback.print_exc();
+            import traceback
+            traceback.print_exc()
             return ''
```
agjohnson: Any reason to catch a generic exception or print traceback here?

autoapi/templates/python/member.rst
```diff
@@ -1,3 +1,5 @@
+.. Member Template
```
agjohnson: This can likely be removed

tests/test_integration.py
```diff
@@ -0,0 +1,93 @@
+import os
+import shutil
+import subprocess as sp
+import unittest
+
+__author__ = 'swenson'
+
+
+class LanguageIntegrationTests(unittest.TestCase):
```
agjohnson: These could be broken out by language. Future testing will likely cover more output feature of the mappers beyond a simple string match. We'll likely want to be testing for child output, some formatting on type output, etc.

agjohnson:
I feel a lot better about what we laid out here, this is a much better abstraction. It feels a lot cleaner and the additional testing makes me feel really good too.

autoapi/mappers/base.py
ericholscher: Yea, it's a .Netism, just didn't want to have to override that whole function in the .Net code. Probably means we need a smaller level of abstraction here, like having a "filename" on the obj.

autoapi/mappers/dotnet.py
ericholscher: This is mainly for debugging. Can probably die now.

tests/test_integration.py
ericholscher: Aye. Definitely just a start.

autoapi/mappers/base.py
agjohnson: Added #12 to deal with this in the future.

agjohnson:
There's some more clean up that can be done here, but we can merge this for now.

PR #11, opened 2015-06-14
Update dotnet codebase for new tooling

ericholscher:
Tests will fail unless you have a proper .Net BuildMeta environment setup. Need to figure out what to do about 3rd party binary deps. Could always just check them into the repo :/


agjohnson:
We should just be mocking the calls to subprocess with StringIO return of the fixture data instead. This way we're testing everything that is part of this library, and making an assumption about what we're getting back from the tools.

autoapi/mappers/dotnet.py
```diff
@@ -1,6 +1,11 @@
+import os
 from collections import defaultdict
+import subprocess
 import yaml
```
agjohnson: `yaml` isn't stdlib, bump import down to third party block of imports

autoapi/mappers/dotnet.py
```diff
@@ -30,7 +48,6 @@ def read_file(self, path, **kwargs):
             with open(path, 'r') as handle:
                 parsed_data = yaml.safe_load(handle)
                 return parsed_data
-        except IOError:
```
agjohnson: Is this intentional?

autoapi/mappers/dotnet.py
```diff
@@ -133,33 +147,32 @@ def _recurse_ns(obj):
...
+            try:
+                filename = obj.name.split('(')[0]
+            except IndexError:
+                filename = id
+            filename = filename.replace('#', '-')
+            detail_dir = os.path.join(root, *filename.split('.'))
+            ensuredir(detail_dir)
+            path = os.path.join(detail_dir, '%s%s' % ('index', source_suffix))
+            with open(path, 'w+') as detail_file:
+                detail_file.write(rst.encode('utf-8'))
```
agjohnson: Catch write IOError here

autoapi/mappers/dotnet.py
```diff
@@ -133,33 +147,32 @@ def _recurse_ns(obj):
...
+
+        # Render Top Index
+        top_level_index = os.path.join(root, 'index.rst')
+        with open(top_level_index, 'w+') as top_level_file:
+            content = self.jinja_env.get_template('index.rst')
+            top_level_file.write(content.render(pages=self.namespaces.values()))
```
agjohnson: Again here, if we need to catch and warn on this. Also could be fine to die and bubble up the error on these blocks I suppose.

ericholscher:
Addressed feedback on both of these PR's. Didn't address testing quite yet, but agree with the need for smaller tidbits. I just didn't want to go down that rabbit hole quite yet.

PR #13, opened 2015-06-23
Integration exp

ericholscher:
This removes dependence on real tooling.
The main downside is that it’s fragile to output changes in the tooling.
Not a lot we can do though,
other than have real integration tests, also.


agjohnson:
Nice, looks good. Excited to fill this up with some tests.

PR #14, opened 2015-07-07
Added the ability to auto-generate a TOCTree entry from AutoAPI.

ericholscher:
This can be turned off with a setting, and only happens if your ``index``
doesn't already have an autoapi block in it.

Also fixes autoapi_file_patterns to take multiple patterns.


autoapi/extension.py
```diff
@@ -41,13 +41,25 @@ def run_autoapi(app):
         'javascript': JavaScriptSphinxMapper,
     }
 
+    default_file_mapping = {
+        'python': ['*.py'],
+        'dotnet': ['project.json', '*.csproj'],
```
agjohnson: Could also add `*.vbproj` to `dotnet` here

autoapi/mappers/dotnet.py
```diff
@@ -20,15 +20,15 @@ class DotNetSphinxMapper(SphinxMapperBase):
...
     top_namespaces = {}
 
-    def load(self, pattern, dir, ignore=[]):
+    def load(self, patterns, dir, ignore=[]):
         '''
         Load objects from the filesystem into the ``paths`` dictionary.
 
         '''
-        for path in self.find_files(pattern='project.json', dir=dir, ignore=ignore):
+        for path in self.find_files(patterns=['project.json'], dir=dir, ignore=ignore):
```
agjohnson: Likewise, reflect the same .NET file patterns here

agjohnson:
LGTM

PR #15, opened 2015-07-12
Changes to support docfx

agjohnson:
Note: This requires a patch on the docfx executable, the hashbang line is
missing, creating an execution failure

This makes the executed command docfx, instead of BuildMeta. It adds some error
checking and reporting to output as well.

This is a work in progress, the output format seems to have changed again.

### DIFF
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -32,7 +32,8 @@ class DotNetSphinxMapper(SphinxMapperBase):
                     stdout=subprocess.PIPE,
                     stderr=subprocess.PIPE,
                     env=dict((key, os.environ[key])
-                              for key in ['PATH', 'DNX_PATH', 'HOME']))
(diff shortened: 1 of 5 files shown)


autoapi/mappers/dotnet.py
```diff
@@ -198,9 +212,9 @@ def __init__(self, obj, **kwargs):
         if syntax is not None:
             # Code example
             try:
-                self.example = syntax['content']['CSharp']
-            except KeyError:
-                pass
+                self.example = syntax['content']
+            except (KeyError, TypeError):
+                traceback.print_exc()
```
ericholscher: Not sure if there's a better pattern here, but I like at least seeing the errors, instead of having them pass silently. Perhaps we should be using the sphinx error logger here, or something.

agjohnson: +1

PR #17, opened 2015-08-03
Add ability for per-mapper cleanup tasks & general docfx cleanup

ericholscher:
This fixes a number of small issues with the configs,
and adds per-mapper cleanup.

This allows the Dotnet mapper to clean up the `_api` directory
and other output that is Mapper specific.

Also concat's all of the files in the call to docfx, which was the main bug with children not appearing correctly in the YAML output. 

I believe these were the last 2 large issues related to local .Net building.

### DIFF
--- a/tests/dotnetexample/README.md
+++ b/tests/dotnetexample/README.md
@@ -1,10 +1,12 @@
 ## .Net example
 
-This assumes that you have a ``MSBuild`` executable on your `PATH`.
+This assumes that you have a ``docfx`` executable on your `PATH`.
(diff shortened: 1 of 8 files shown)


tests/dotnetexample/README.md
```diff
@@ -1,10 +1,12 @@
 ## .Net example
```
agjohnson: Just a note on usage here, it doesn't affect this PR: `.Net` should be `.NET` wherever we use it

agjohnson:
+1 besides addressing failing tests.

PR #18, opened 2015-08-03
Add tox, which adds linting and doc building to testing

ericholscher:
We will have a few failures on linting,
so I haven't turned this on for Travis specifically yet.


agjohnson:
LGTM

PR #19, opened 2015-08-03
Fix up python 3 support.

ericholscher:


tests/test_domains.py
```diff
@@ -67,29 +66,27 @@ def _create_class(data):
         self.assertIsInstance(cls, dotnet.DotNetClass)
         self.assertDictEqual(cls.item_map, {})
 
+    def _mock_find(self, patterns, **kwargs):
+        return {'items': ['foo', 'bar']}
+
+    def _mock_read(self, path):
+        return {'items': [{'id': 'Foo.Bar', 'name': 'Foo', 'type': 'property'},
+                          {'id': 'Foo.Bar2', 'name': 'Bar', 'type': 'property'}],
+                'id': 'Foo.Bar', 'type': 'Class', 'summary': path}
```
agjohnson: Should probably just define these methods as standard mock objects instead of moving them out of the test scope they are established for

autoapi/mappers/python.py
```diff
@@ -19,8 +18,7 @@ def read_file(self, path, **kwargs):
 
         :param path: Path of file to read
         '''
-        # TODO support JSON here
-        # TODO sphinx way of reporting errors in logs?
+        from epyparse import parsed
```
agjohnson: A more expected pattern here would be fencing the import in the module scope with a `sys.version_info < (3,)` check

agjohnson:
There are some py3/py26 incompatibilities here we could address to make the module usable on py26. We can just address that separately if we feel we should.

PR #20, opened 2015-08-11
Add Windows builds

ericholscher:
This adds appveyor support, and builds for Windows. Currently they are failing on Python 3 because of an error in the Babel library.

### DIFF
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -36,7 +36,7 @@ class DotNetSphinxMapper(SphinxMapperBase):
             all_files.add(_file)
         if all_files:
             try:
-                command = ['docfx', 'metadata', '--raw', '--force']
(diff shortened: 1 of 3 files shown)


autoapi/mappers/dotnet.py
```diff
@@ -180,16 +180,16 @@ def output_rst(self, root, source_suffix):
                 filename = obj.name.split('(')[0]
             except IndexError:
                 filename = id
-            filename = filename.replace('#', '-')
+            filename = filename.replace('#', '-').replace('<', '-').replace('>', '')
```
ericholscher: We should probably just be properly slugifying these. I made the < turn into a -, but killing the trailing >, but we should standardize on a schema here.

agjohnson: Agreed. A simpler approach would be a regex with non-match replacements to catch everything.

agjohnson:
Opened #21 to address slugging later, LGTM otherwise.

PR #22, opened 2015-08-17
Stub out XML parsing

ericholscher:
Adds XML doc comment parsing on description, parameter, and return fields. It handles the following tags:

* `see` and `seealso` - replaces with a reference link
* `paramref` and `typeparamref` - just replaces with bold text as we can't reference parameters


agjohnson:
Failing tests are a symptom of the failing slugging, fixed in #23

PR #23, opened 2015-08-19
Use more proper slugging on filenames for object

### LINKED ISSUE #21: Improve slugging
As per discussion in #20, make proper slugging on filename output.

agjohnson:
This resolves some issues with special characters in paths on Windows based
systems. It replaces special characters and unicode characters to ensure paths
are predictable.

Fixes #21

### DIFF
--- a/autoapi/mappers/base.py
+++ b/autoapi/mappers/base.py
@@ -1,7 +1,9 @@
+import re
 import os
 import fnmatch
 from collections import OrderedDict
 
+import unidecode
 from jinja2 import Environment, FileSystemLoader, TemplateNotFound
 from sphinx.util.console import darkgreen, bold
 from sphinx.util.osutil import ensuredir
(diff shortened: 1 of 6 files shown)


autoapi/mappers/base.py
```diff
@@ -1,7 +1,9 @@
+import re
 import os
 import fnmatch
 from collections import OrderedDict
 
+import unidecode
```
ericholscher: Does this need to be added to our reqs?

setup.py
```diff
@@ -9,6 +9,7 @@
             'sphinx',
             'sphinxcontrib-golangdomain',
             'sphinxcontrib-dotnetdomain',
+            'unidecode',
```
ericholscher: We can probably just switch to using setuptools

autoapi/mappers/base.py
```diff
@@ -96,6 +98,27 @@ def short_name(self):
...
+        * Substitute hyphens for non-word characters
+        * Break up the string as paths
+        '''
+        slug = self.name
+        try:
+            slug = self.name.split('(')[0]
+        except IndexError:
+            pass
+        slug = unidecode.unidecode(slug)
+        slug = slug.replace('-', '')
```
ericholscher: Is this to remove collision possibilities? It seems like we might want to keep them around.

agjohnson: I suppose The Right Thing To Do would be to track and detect duplicates here. Each step of slugging introduces an area that could conflict with another path. Removing dashes wouldn't be important in that case.

This might have to move to the mapper afterall, so that duplicates across object names can be tracked there.

ericholscher:
Added #25 to address setuptools/distutils.

agjohnson:
I opened #26 to address slugging deduplication as well.

PR #27, opened 2015-08-21
Fixes necessary for running on Windows

agjohnson:
Previously, we were expecting the command would always be `docfx`, which on a windows box, the command is `docfx.cmd`. This configures the command based on system and also removes a problematic subprocess `shell=True` argument.

The `shell=True` argument wouldn't be needed, except the `docfx` tool on linux is missing it's shebang line and this proxy command fails if subprocess isn't using shell emulation.

This will be blocking until we have a fix for removing `shell=True` for unix systems.

### DIFF
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -1,5 +1,6 @@
 import re
 import os
+import platform
 import subprocess
 import traceback
 import shutil
@@ -12,6 +13,10 @@ from sphinx.errors import ExtensionError
 
 from .base import PythonMapperBase, SphinxMapperBase
 
+
+DOCFX_COMMAND = 'docfx'
+if platform.system() == 'Windows':
+    DOCFX_COMMAND = 'docfx.cmd'
 
 # Doc comment patterns
 DOC_COMMENT_PATTERN = r'''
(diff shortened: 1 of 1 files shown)


danroth27:
I was messing around with this as well and it looks like calling `docfx` works if `shell=True` is specified. If `shell=False` then you need to specify `docfx.cmd`. I think it's preferred to not take a dependency on DNX using a .cmd file - conceptually you should be able to install a DNX command and then call it without any specific file extension. Unfortunately, this does mean continuing to use `shell=True` which has it's own [security risks](https://docs.python.org/2.7/library/subprocess.html?highlight=subprocess#frequently-used-arguments).

However, the custom env dictionary seems to be getting in the way in my tests. I'm not sure why it doesn't work, but if I remove the env parameter then things work fine. Is this parameter really needed? Isn't inheriting the environment of the current process sufficient?

It also seems preferred to pass a sequence of [args](https://docs.python.org/2.7/library/subprocess.html#frequently-used-arguments) instead of a string as "it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names)."

Alternative PR: https://github.com/rtfd/sphinx-autoapi/pull/28

agjohnson:
I see now, requiring `docfx.cmd` was part of my testing. I was testing with mingw32 + bash and so the full name of the command was expected. Running outside mingw32 now I get the behavior you're noticing, and your fix does resolve these issues. This means we're missing something from the environment variable pruning.

The output I get from docfx when I do prune the env variables is: `Failed to initialize runtime (80090006)`

agjohnson:
My last commit should address the env variables that were missing, causing docfx to fail. I got passing builds with both this and #28.

danroth27:
That seems to work for me.

PR #40, opened 2015-10-27
Fix template path issues on windows

### LINKED ISSUE #37: .NET toctrees use unix path separators in templates
TOC tree includes in templates should determine the correct separator to use.

agjohnson:
Replaces OS path separators with unix style separators, since that's what
Sphinx is expecting in toctree directives.

Fixes #37

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -12,6 +12,7 @@ from sphinx.addnodes import toctree
 from sphinx.errors import ExtensionError
 
 from .backends import default_file_mapping, default_ignore_patterns, default_backend_mapping
+from .settings import URL_ROOT
 
 default_options = ['members', 'undoc-members', 'private-members', 'special-members']
 
@@ -48,7 +49,8 @@ def run_autoapi(app):
     app.env.autoapi_data = []
 
     domain = default_backend_mapping[app.config.autoapi_type]
-    domain_obj = domain(app, template_dir=app.config.autoapi_template_dir)
+    domain_obj = domain(app, template_dir=app.config.autoapi_template_dir,
+                        url_root=app.config.autoapi_url_root)
 
     if app.config.autoapi_file_patterns:
         file_patterns = app.config.autoapi_file_patterns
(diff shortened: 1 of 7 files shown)


autoapi/mappers/dotnet.py
```diff
@@ -355,6 +355,14 @@ def ref_short_name(self):
         '''Same as above, return the truncated name instead'''
         return self.ref_name.split('.')[-1]
 
+    @property
+    def include_path(self):
+        """Return 'absolute' path without regarding OS path separator"""
+        parts = ['/autoapi']
```
ericholscher: This should probably come from the setting that sets the path of the output, which won't necessarily always be `autoapi`

autoapi/extension.py
```diff
@@ -124,4 +126,5 @@ def setup(app):
     app.add_config_value('autoapi_keep_files', False, 'html')
     app.add_config_value('autoapi_add_toctree_entry', True, 'html')
     app.add_config_value('autoapi_template_dir', [], 'html')
+    app.add_config_value('autoapi_url_root', URL_ROOT, 'html')
```
ericholscher: isn't this just `autoapi_root` already defined?

ericholscher:
This doesn't need a new setting, it just needs to respect `autoapi_root`, which is already set for where we output files.

autoapi/extension.py
agjohnson: It wasn't clear based on the docs, and it seemed it could be an absolute path based on tests. If the intended use of `autoapi_root` was as a path name inside the local path, we should clarify the docs and rework this setting.

ericholscher: It's relative in every place: http://sphinx-autoapi.readthedocs.org/en/latest/config.html#confval-autoapi_root & https://github.com/rtfd/sphinx-autoapi/blob/master/autoapi/extension.py#L119

It gets normalized just above the code you edited: https://github.com/rtfd/sphinx-autoapi/blob/master/autoapi/extension.py#L46

ericholscher:
:+1:

PR #42, opened 2015-11-08
Fix up generated path separator when on Windows

danroth27:
On windows os.path.join will generate a path with forward slashes, which messes up the generated toctree paths. 

This little hack will fix up the path separators to be forward slashes. There's probably a more correct way to do this, but this seems to works for most cases.

### DIFF
--- a/autoapi/mappers/base.py
+++ b/autoapi/mappers/base.py
@@ -122,7 +122,7 @@ class PythonMapperBase(object):
         slug = unidecode.unidecode(slug)
         slug = slug.replace('-', '')
         slug = re.sub(r'[^\w\.]+', '-', slug).strip('-')
-        return os.path.join(*slug.split('.'))
+        return os.path.join(*slug.split('.')).replace('\\', '/')
 
     @property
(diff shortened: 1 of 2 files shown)


autoapi/mappers/base.py
```diff
@@ -122,7 +122,7 @@ def pathname(self):
         slug = unidecode.unidecode(slug)
         slug = slug.replace('-', '')
         slug = re.sub(r'[^\w\.]+', '-', slug).strip('-')
-        return os.path.join(*slug.split('.'))
+        return os.path.join(*slug.split('.')).replace('\\', '/')
```
guillermooo: Assuming the string doesn't contain other separators, maybe this is simpler:

```python
# Paths must be Unix-style, even on Windows.
'/'.join(slug.split('.'))
```

agjohnson:
This should be addressed with #40. Another property was added there, as the pathname is used outside toctree inclusion.

ericholscher:
Closing old PR, this was already fixed.

PR #51, opened 2016-01-13
.NET template output breaks definition lists on listing pages

agjohnson:
See rtfd/sphinxcontrib-dotnetdomain#45 for more info.

screen shot 2016-01-12 at 4 04 13 pm


danroth27:
Nice!

PR #66, opened 2016-02-25
Add resolution to spec identifiers using docfx references

### LINKED ISSUE #58: Handling generic types
Generic types in the docfx metadata format are identified using a [spec identifier](http://dotnet.github.io/docfx/spec/metadata_dotnet_spec.html#2-2-spec-identifiers) format that represents generic type parameters with curly brackets. This format is for referencing purposes in the metadata documents, but it is showing up in the rendered docs (ex see the user argument and the return value of the [GetSecurityStampAsync](https://docs.asp.net/projects/api/en/latest/autoapi/Microsoft/AspNet/Identity/EntityFramework/UserStore-TUser-TRole-TContext-TKey/index.html?highlight=getsecuritystampasync#Microsoft.AspNet.Identity.EntityFramework.UserStore<TUser, TRole, TContext, TKey>.GetSecurityStampAsync) method:

image

image

Instead these "spec identifiers" should be resolved to their user friendly names for display purposes. I believe this data is available for lookup in the references section of the generated metadata documents:

```yaml
- uid: System.Threading.Tasks.Task{System.String}
  parent: System.Threading.Tasks
  definition: System.Threading.Tasks.Task`1
  name: Task<String>
  name.vb: Task(Of String)
  fullName: System.Threading.Tasks.Task<System.String>
  fullName.vb: System.Threading.Tasks.Task(Of System.String)
```

agjohnson:
This uses the docfx output references to resolve the spec identifier to a human
friendly name. If the reference is a nested reference, build up output for parameter and return defined types, which the domain will deconstruct and link separately. This also fixes `paramref` xml parsing and display.

The parameter field output uses a special syntax that the domain will parse out, from rtfd/sphinx-contribdotnetdomain#54. The output looks like:

```
    :type foo: Foo<Foo`1>{Bar<Bar`1>{TUser}}
```

This provides two references that will resolve, both with aliased targets, and one, `TUser` which will ultimately fail resolution

Example:
screen shot 2016-03-04 at 10 44 01 am

This work also adds hidden obj definitions so that both reference syntaxes are linkable:

```
:dn:cls:`Foo<TBar>`
:dn:cls:`Foo\`1`
```

Fixes #58
Requires rtfd/sphinxcontrib-dotnetdomain#54

Work left:
- [x] Resolve each identifier in strings like `Foo<Bar<Task>>` -- it might make sense to do this as a more general update to the dotnet domain reference handling.

### DIFF
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -276,8 +276,10 @@ class DotNetPythonMapper(PythonMapperBase):
                             param.get('description', ''))
                     })
 
-            self.returns = self.transform_doc_comments(
-                syntax.get('return', None))
+            self.returns = {}
+            self.returns['type'] = syntax.get('return', {}).get('type')
+            self.returns['description'] = self.transform_doc_comments(
+                syntax.get('return', {}).get('description'))
 
         # Inheritance
         # TODO Support more than just a class type here, should support enum/etc
@@ -400,8 +402,19 @@ class DotNetPythonMapper(PythonMapperBase):
                 text_start = re.sub(r'(\S)$', r'\1 ', text_start)
 
                 text = ''.join([text_start, replacement, text_end])
-            text = DOC_COMMENT_PARAM_PATTERN.sub(
-                '``\g<attr_value>``', text)
+            while True:
+                found = DOC_COMMENT_PARAM_PATTERN.search(text)
+                if found is None:
+                    break
+
+                # Escape following text
+                text_end = text[found.end():]
+                text_start = text[:found.start()]
+                text_end = re.sub(r'^(\S)', r'\\\1', text_end)
+                text_start = re.sub(r'(\S)$', r'\1 ', text_start)
+
+                text = ''.join([text_start, '``', found.group('attr_value'),
+                                '``', text_end])
         except TypeError:
             pass
         return text
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -112,8 +112,10 @@ class DotNetSphinxMapper(SphinxMapperBase):
     def map(self, options=None, **kwargs):
         '''Trigger find of serialized sources and build objects'''
         for path, data in self.paths.items():
+            references = data['references']
             for item in data['items']:
-                for obj in self.create_class(item, options):
+                for obj in self.create_class(item, options,
+                                             references=references):
                     self.add_object(obj)
 
         self.organize_objects()
@@ -144,7 +146,7 @@ class DotNetSphinxMapper(SphinxMapperBase):
             self.app.warn('Unknown type: %s' % data)
         else:
             obj = cls(data, jinja_env=self.jinja_env, options=options,
-                      url_root=self.url_root)
+                      url_root=self.url_root, **kwargs)
 
             # Append child objects
             # TODO this should recurse in the case we're getting back more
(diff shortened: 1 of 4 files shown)


agjohnson:
Okay, so I did in fact unearth more questions on this implementation. Cases like the following work as expected:

screen shot 2016-02-26 at 4 21 07 pm

This is, `Microsoft.AspNet.Identity.UserManager<TUser>` links to the correct class definition, as there is a `Microsoft.AspNet.Identity.UserManager<TUser>` defined as `TUser` is the named parameter. This seems it should stay the same?

However, if the generic is defined as a type class, a reference such as:
`Microsoft.Extensions.OptionsModel.IOptions<Microsoft.AspNet.Identity.DataProtectionTokenProviderOptions>`

Could actually be defined as something such as:
```
Microsoft.Extensions.OptionsModel.IOptions<TOptions>
```

So, it seems we should rather be linking to:
```
Microsoft.Extensions.OptionsModel.IOptions`1
```

The docfx output does give me `definition` field on in *some* of these cases, that is:
```yaml
- uid: Microsoft.Extensions.OptionsModel.IOptions{Microsoft.AspNet.Identity.DataProtectionTokenProviderOptions}
  parent: Microsoft.Extensions.OptionsModel
  definition: Microsoft.Extensions.OptionsModel.IOptions`1
  href: Microsoft.AspNet.Identity.DataProtectionTokenProviderOptions.yml
  name: IOptions<DataProtectionTokenProviderOptions>
  name.vb: IOptions(Of DataProtectionTokenProviderOptions)
  fullName: Microsoft.Extensions.Opti
```

Does the problem with linking make sense? Is my understanding of this far off?

agjohnson:
cc @danroth27 @shirhatti

danroth27:
@vicancy Can you help out with some of these questions as well? We want to make sure we are interpreting the docfx metadata format correctly.

vicancy:
DocFX uses `spec<.language>` property to define every part of the value.
`Microsoft.Extensions.OptionsModel.IOptions<Microsoft.AspNet.Identity.DataProtectionTokenProviderOptions>` should contain 2 links, one for ``Microsoft.Extensions.OptionsModel.IOptions`1`` and another for `Microsoft.AspNet.Identity.DataProtectionTokenProviderOptions`.

An example `System.Threading.Tasks.Task{System.Net.Http.HttpResponseMessage}` as defined in https://github.com/docascode/coreapi/blob/master/System.Net.Http.HttpMessageInvoker.yml#L180:
```yaml
- uid: System.Threading.Tasks.Task{System.Net.Http.HttpResponseMessage}
  definition: System.Threading.Tasks.Task`1
  name: Threading.Tasks.Task<HttpResponseMessage>
  name.vb: Threading.Tasks.Task(Of HttpResponseMessage)
  fullName: System.Threading.Tasks.Task<System.Net.Http.HttpResponseMessage>
  fullName.vb: System.Threading.Tasks.Task(Of System.Net.Http.HttpResponseMessage)
  spec.csharp:
  - uid: System.Threading.Tasks.Task`1
    name: Threading.Tasks.Task
    fullName: System.Threading.Tasks.Task
    isExternal: true
  - name: <
    fullName: <
  - uid: System.Net.Http.HttpResponseMessage
    name: HttpResponseMessage
    fullName: System.Net.Http.HttpResponseMessage
  - name: '>'
    fullName: '>'
  spec.vb:
  - uid: System.Threading.Tasks.Task`1
    name: Threading.Tasks.Task
    fullName: System.Threading.Tasks.Task
    isExternal: true
  - name: '(Of '
    fullName: '(Of '
  - uid: System.Net.Http.HttpResponseMessage
    name: HttpResponseMessage
    fullName: System.Net.Http.HttpResponseMessage
  - name: )
    fullName: )
```

And it is rendered as the return type for `SendAsync(HttpRequestMessage, Threading.CancellationToken)` in http://dotnet.github.io/api/System.Net.Http.HttpMessageInvoker.html#System_Net_Http_HttpMessageInvoker_SendAsync_System_Net_Http_HttpRequestMessage_System_Threading_CancellationToken_

agjohnson:
Thanks @vicancy!

This was close to what I was planning to do, but just wasn't sure how to handle the linking for a case where the definition is actually defined. That is, a reference such as `Foo<TBar>` will already resolve, but a nested reference like `Foo<System.String>` would not have -- this would instead be linked to with ```Foo`1```.

I've updated the description to reflect my changes, including an example of the output now.

ericholscher:
:+1:

Issue #53, opened 2016-02-03
No autoapi_dir config value


danroth27:
When I try to use the autoapi extension I get this error complaining that there is no autoapi_dir config value specified:

```
# Sphinx version: 1.3.1
# Python version: 3.5.1 (CPython)
# Docutils version: 0.12 release
# Jinja2 version: 2.8
# Last messages:

# Loaded extensions:
Traceback (most recent call last):
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\sphinx\cmdline.py", line 244, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\sphinx\application.py", line 188, in __init__
    self._init_builder(buildername)
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\sphinx\application.py", line 250, in _init_builder
    self.emit('builder-inited')
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\sphinx\application.py", line 497, in emit
    results.append(callback(self, *args))
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\autoapi\extension.py", line 32, in run_autoapi
    normalized_dirs.append(app.config.autoapi_dir)
  File "c:\users\daroth\appdata\local\programs\python\python35\lib\site-packages\sphinx\config.py", line 368, in __getattr__
    raise AttributeError('No such config value: %s' % name)
AttributeError: No such config value: autoapi_dir
```

This code in extension.py appears to be the problem:

```python
    # Make sure the paths are full
    normalized_dirs = []
    for path in app.config.autoapi_dirs:
        if os.path.isabs(path):
            normalized_dirs.append(app.config.autoapi_dir)
        else:
            normalized_dirs.append(
                os.path.normpath(os.path.join(app.confdir, path))
            )
```

I think `app.config.autoapi_dir` should be `path`.

PR #70, opened 2016-05-02
Handle string input for api dir

agjohnson:
This was iterating over the settings assuming it was a list. We handle this as a
list in some instances, and recommend using a string in others. Just handle
both.

Refs #53


agjohnson:
cc xer0x -- I believe you mentioned hitting this issue on irc

ericholscher:
:+1:

PR #71, opened 2016-05-02
Use short name for reference link name

### LINKED ISSUE #63: Types listed on a namespace page should note include the namespace in the type name
Currently the generated namespace pages list the available types in that namespace using the full name of the types (including the namespace). 

image

This is very redundant as the namespace is indicated at the top of the page. Instead the types should be listed using only the local name of the type without the namespace.

agjohnson:
As the parent block is a namespace declaration, we can use nesting to imply the
namespace. This removes the namespace from the object name and matches what we
do on other pages.

Fixes #63


ericholscher:
👍

Issue #60, opened 2016-02-03
Generated .NET API docs for a type should include the namespace and assembly name


danroth27:
Existing MSDN docs for .NET types include the namespace and assembly name near the top of the page. For example https://msdn.microsoft.com/en-us/library/system.console(v=vs.110).aspx:

![image](https://cloud.githubusercontent.com/assets/1874516/12797743/54fc6d92-ca7a-11e5-8644-298c4b4bb34e.png)

It would be great if autoapi did this as well. I believe the metadata is already provided by docfx.

agjohnson:
Looking at the generated data, the assembly names for `Identity` are `Microsoft.AspNet.Identity`. It seems we can't assume this will be an object with an id to be referenced?

danroth27:
@vacancy @chenkennt Can you guys point @agjohnson at the right metadata to use?

chenkennt:
Namespace is a reference to another object, assembly is just a plain text.

agjohnson:
Great, I'm treating it as such currently, assuming there will be a string that can't be referenced eventually. Sounds like this is the way to go.

agjohnson:
Closing this for now, it might make sense to dedupe this display eventually, in the case that the assembly and namespace are matching/etc.

PR #72, opened 2016-05-02
Clean up .NET detail page, add assemblies list

agjohnson:
The .NET detail page was slightly messy, this cleans up and settles on some UX
there:

* Don't use an explicit header for summary, move it up under the object heading
* Drop redundant (and broken) github edit link on page, we need to fix the
  github header link instead.
* Add pattern for displaying read only fields in definition list
* Adds namespace + assemblies listing to detail page

Refs #60

screen shot 2016-05-01 at 7 57 14 pm


ericholscher:
:+1:

PR #78, opened 2016-06-09
Start using pydocstyle for Python doc generation

ericholscher:
This adds the ability to use pydocstyle for both Python 2 and 3 docstring extraction. Still a Work in Progress.


agjohnson:
I pushed up some light fixes to the this PR, but didn't address any of the large issues I see with this.

I opened #80, #81, #82, #83 to address problems introduced here. I'd prefer to merge and start working through some of these issues, as this has been static for a while now.

autoapi/mappers/base.py
```diff
@@ -47,8 +49,9 @@ class PythonMapperBase(object):
     # Create a page in the output for this object.
     top_level_object = False
 
-    def __init__(self, obj, options=None, jinja_env=None, url_root=None):
+    def __init__(self, obj, path, options=None, jinja_env=None, url_root=None):
```
agjohnson: Any reason this was added on the base python mapper class? This breaks all of our implementations. Seems it could just live on the Python mapper.

PR #87, opened 2016-11-03
Build initial toctree insertion implementation

ericholscher:


autoapi/toctree.py
```diff
@@ -0,0 +1,141 @@
...
+    for desc_node in doctree.traverse(addnodes.desc):
+        try:
+            ref_id = desc_node.children[0].attributes['ids'][0]
+        except IndexError as e:
+            print('Invalid desc node: %s' % e)
+            continue
+        try:
+            # Python domain object
+            ref_text = desc_node[0].attributes['fullname'].split('.')[-1].split('(')[0]
+        except:
```
agjohnson: Bare except here, replace with `KeyError`/`IndexError`

autoapi/toctree.py
```diff
@@ -0,0 +1,141 @@
...
+            # Use `astext` for other types of domain objects
+            ref_text = desc_node[0].astext().split('.')[-1].split('(')[0]
```
agjohnson: Perhaps `astext` is the wrong way here. Are there child nodes in `desc_node` that we could grab from -- that is, grab the last child here? I'd mark this for future work.

tests/test_integration.py
```diff
@@ -134,3 +134,13 @@ def test_toctree_overrides(self):
             '_build/text/index.txt',
             'AutoAPI Index'
         )
+
+    def test_toctree_domain_insertion(self):
+        """
+        Test that the example_function gets added to the TOC Tree
+        """
+        self._run_test(
+            'toctreeexample',
```
agjohnson: This path is missing

autoapi/toctree.py
```diff
@@ -0,0 +1,141 @@
...
+    Find the actual TOC node for a ref_id.
+    Depends on the object type:
+
+    * Section - First section (refuri) or 2nd+ level section (anchorname)
+    * Desc - Just use the anchor name
+    """
+    for check_node in toc.traverse(nodes.reference):
+        if objtype == nodes.section and \
+                (check_node.attributes['refuri'] == ref_id or
+                 check_node.attributes['anchorname'] == '#' + ref_id):
```
ericholscher: Don't believe so, but not 100% sure. Will leave for now and see if it blows up on larger use.

autoapi/mappers/python.py
```diff
@@ -174,8 +173,9 @@ def _get_arguments(obj):
         # exceptions, including SyntaxError
         try:
             parsed = ast.parse(source)
-        except:  # noqa
-            return
+        except Exception as e:  # noqa
+            print("Error parsing AST: %s" % str(e))
```
ericholscher: Don't have access to Sphinx app, so leaving this for now.

tests/test_integration.py
ericholscher: It's piggy backing on existing test.

PR #93, opened 2017-02-02
Fixed exception when parsing unpacked Python arguments

### LINKED ISSUE #92: AttributeError on Python functions with unpacked arguments
When the PythonSphinxMapper._get_arguments is called with a an object representing a function with unpacked arguments (eg `def test(one, (two, three), four):`), autoapi fails with:

```
  File "venv/lib/python2.7/site-packages/autoapi/mappers/python.py", line 181, in _get_arguments
    for arg in parsed_args.args]
AttributeError: 'Tuple' object has no attribute 'id'
```

We need to change how the `arg_names` list is generated so that a single item in `parsed_args.args` (specifically a `Tuple`) can add multiple names to the list.

AWhetter:
Fixes #92

### DIFF
--- a/autoapi/mappers/python.py
+++ b/autoapi/mappers/python.py
@@ -167,6 +167,23 @@ class PythonPythonMapper(PythonMapperBase):
         :returns: list of argument or argument and value pairs
         :rtype: list
         """
+        def get_arg_names(args):
+            """Get the names of each ast-parsed argument in the given list.
+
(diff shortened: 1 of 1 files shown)


AWhetter:
Apparently tuple unpacking of parameters has been removed in Python 3 so it doesn't really make sense to need to support this. Closing and I'll close the issue as well.

PR #94, opened 2017-02-02
Fix for Sphinx 1.5

AWhetter:
`build_toc_from` has been moved to `Toctree.process_doc` in Sphinx 1.5


ericholscher:
Does this break on Sphinx <1.5?

AWhetter:
Yes it will.

ericholscher:
Great, thanks!

agjohnson:
This should instead check for the Sphinx version and apply this workaround conditionally, instead of pinned to `1.5+` for this fix.

ericholscher:
Fixed in #95

PR #95, opened 2017-02-08
Add fix to version checking on <1.5

ericholscher:


autoapi/extension.py
```diff
@@ -134,7 +135,10 @@ def doctree_read(app, doctree):
             app.info(bold('[AutoAPI] ') +
                      darkgreen('Adding AutoAPI TOCTree [%s] to index.rst' % toc_entry)
                      )
-            app.env.toctree.process_doc(app.env.docname, doctree)
+            if sphinx.version_info[0] >= 1 and sphinx.version_info[1] >= 5:
```
agjohnson: A more succinct way of checking this is:

```python
if sphinx.version_info >= (1, 5):
```

AWhetter: Agreed. At the moment the code will fail for sphinx "2.0" for example because the 0 in `sphinx.version_info[1]` won't be >= 5.

Issue #99, opened 2017-02-08
Expand ast parsing to allow for inheritance docs, data docs, and class signatures


AWhetter:
This is a pretty hefty issue that will probably need breaking down if it's decided that it's a good idea, but changing how Python files are parsed should bring a number of benefits:

1. Inheritance documentation: autodoc is able to list what a class inherits from and link through to that class' documentation. As far as I know, pydocstyle isn't able to determine this information. I find this to be an incredibly useful feature.
2. Data documentation: It's valid to have a module level variable and an associated docstring. I'm not actually sure if pydocstyle picks this up, but currently they aren't documented by autoapi. autodoc can find these docstring with the `autodata` directive.
3. Class signature: autodoc can display the arguments given to the constructor of a class in the class signature. This isn't always possible with parsing (eg maybe the base class exists in a file that isn't being parsed), but not using pydocstyle should allow us to pick this up at least some of the time.

agjohnson:
I looked into using astroid, and maybe it was a matter of the docs being non-existent, but this library doesn't do what we need it to. I did try some other approaches, and tried to find something that could be extended, if anything.

I think this comes back to either writing a custom AST parser, or making pydocstyle do what we need. While pydocstyle isn't optimal, or even really written in a way that makes it easy to customize, I'd rather not introduce Yet Another Parser.

I started some of this in #104, I hope to add more for inheritance and the module/class variable documentaiton pattern that `autodoc` exposes. For now, this allows us to build more projects with strange patterns for defining `__all__`.

AWhetter:
Nice! Thanks for looking into this. I didn't know how easy it would be to override thepydocstyle parser, and I agree that it seems like a much less radical option than astroid.
Really this issue is more about the inheritance, data, and class signature documentation rather than what ast parsing method we use so I'll change the title to reflect that.

AWhetter:
I've prototyped an ast parser because after seeing what we had to do to override the pydocstyle parser, I thought that I could come up with something simpler.

I've put this prototype here: https://github.com/AWhetter/sphinx-autoapi/compare/py_hierarchy...AWhetter:fix_99

It's fewer lines of code but more importantly I think that it's easier to understand. It is using astroid, mostly because I know it, but I did start converting it to use the ast module instead. However ast would require way more code because you don't get much for free. Have a look and let me know what you think.

agjohnson:
[...]

PR #104, opened 2017-04-07
Start altering the pydocstyle AST parser for our needs

agjohnson:
This forks some of the pydocstyle AST parser into out mapper. Eventually, some
of the other operations such as parsing arguments and performing full name
lookup can be moved in as well. For now, this is not doing any extra assignment
tracking/etc, several of these operations will just throw this information out.

Refs #99

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,20 @@
 language: python
 python:
- - 2.7
+  - 2.7
+  - 3.4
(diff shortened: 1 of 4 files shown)


ericholscher:
Should we update the docs here too?

agjohnson:
Yeah, i'll put up another PR with doc updates.

PR #106, opened 2017-04-07
Fix Python argument parsing for Call nodes

agjohnson:
This previous was throwing exceptions in some cases where Attribute nodes were
used instead of Name nodes.


ericholscher:
:+1:

PR #113, opened 2017-06-28
Basic cleanup tasks

### LINKED ISSUE #98: Properly format Google and Numpy (napoleon) docstrings in Python
Currently Google docstrings are inserted into the rendered rst directly, instead of being parsed by Sphinx's napoleon extension.
Sphinx implements napoleon via autodoc events. Part of fixing this issue might be to make changes to napoleon to make it easier to use outside of autodoc, though I'm pretty sure we should be able to use (https://github.com/sphinx-doc/sphinx/blob/6fa0262802a09050e09445c9fd630c69b5ad1204/sphinx/ext/napoleon/__init__.py#L326), or at least mimic it seeing as it's protected.

ericholscher:
This PR includes:

*    Add ability to add API to top-level TOC.
*    Fix templates to nest functions and use short names

Fixes #98 
Fixes rtfd/pydoc.io#17
Fixes rtfd/pydoc.io#18

### DIFF
--- a/autoapi/mappers/base.py
+++ b/autoapi/mappers/base.py
@@ -5,6 +5,7 @@ from collections import OrderedDict, namedtuple
 
 import unidecode
 from jinja2 import Environment, FileSystemLoader, TemplateNotFound
+import sphinx.util
 from sphinx.util.console import darkgreen, bold
 from sphinx.util.osutil import ensuredir
 from sphinx.util.docstrings import prepare_docstring
@@ -199,7 +200,7 @@ class SphinxMapperBase(object):
         # Mapping of {namespace id -> Python Object}
         self.top_level_objects = OrderedDict()
 
-    def load(self, patterns, dirs, ignore=None, **kwargs):
+    def load(self, patterns, dirs, ignore=None):
         '''
         Load objects from the filesystem into the ``paths`` dictionary.
 
@@ -238,7 +239,12 @@ class SphinxMapperBase(object):
 
                         files_to_read.append(filename)
 
-        for _path in self.app.status_iterator(
+        if sphinx.version_info >= (1, 6):
+            status_iterator = sphinx.util.status_iterator
+        else:
+            status_iterator = self.app.status_iterator
+
+        for _path in status_iterator(
                 files_to_read,
                 '[AutoAPI] Reading files... ',
                 darkgreen,
@@ -272,7 +278,7 @@ class SphinxMapperBase(object):
         '''
         Create class object.
 
-        :param obj: Instance of a AutoAPI object
+        :param data: Instance of a AutoAPI object
         '''
         raise NotImplementedError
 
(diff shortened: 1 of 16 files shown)


autoapi/templates/python/function.rst
```diff
@@ -1,6 +1,6 @@
 {%- if obj.display %}
 
-.. function:: {{ obj.name }}({{ obj.args|join(',') }})
+.. function:: {{ obj.short_name }}({{ obj.args|join(',') }})
```
agjohnson: We should ensure linking is correct with this change. We are not currently testing whether this is nested correctly.

prospector.yml
```diff
@@ -15,6 +15,7 @@ pylint:
   max-line-length: 100
   disable:
       - interface-not-implemented
+      - arguments-differ
```
agjohnson: We shouldn't blanket turn this on.

tox.ini
```diff
@@ -1,9 +1,16 @@
 [tox]
 envlist =
-    py{27,34,35}-sphinx{13,14,15}
+    py{27,34,35,36}-sphinx{13,14,15}
```
agjohnson: The original change just dropped 13 due to failing tests, we should instead ensure 1.3 support unless we are dictating which versions are deprecated. As far as I can tell, there is no reason not to support 1.3, it's just that state is stored in a different place.

autoapi/templates/python/function.rst
ericholscher: What in particular are you worried about? I tested it at the module level and inside a class.

tox.ini
ericholscher: It's a super old version of Sphinx. I'd rather ask what the value is of supporting 1.3?

ericholscher: I can't find an actual version range of Sphinx that they support, but I'd guess 1.3 isn't officially supported anymore.

autoapi/templates/python/function.rst
agjohnson: That the nesting has changed or is not happening and therefore relative ref linking doesn't work. This is currently a blind spot in the tests.

tox.ini
agjohnson: Yeah, 1.3 is old, but folks are still using it on RTD. I'd be curious to know stats around it. Anyways, it's easy enough to support both, the env state that it is failing on was moved (i thought in 1.5, but apparently 1.4)

autoapi/mappers/base.py
```diff
@@ -180,7 +181,10 @@ def __init__(self, app, template_dir=None, url_root=None):
         )
 
         def _wrapped_prepare(value):
-            return '\n'.join(prepare_docstring(value))
+            preped_value = '\n'.join(prepare_docstring(value))
+            numpy_value = str(NumpyDocstring(preped_value))
+            final_value = str(GoogleDocstring(numpy_value))
```
AWhetter: Google and Numpy docstrings are a Python only thing, so it seems weird to put this here. Also this won't work for the cases where the summary line is not indented. We'd need to pass it into `prepare_docstring` to handle this case.
```
"""This won't work
...
"""
This will
```

ericholscher:
This should be ready for re-review.

PR #114, opened 2017-08-28
Fix for Sphinx 1.6

AWhetter:
This fixes autoapi for Sphinx 1.6. We actually don't need to call `process_doc` at all. `process_doc` edits the doctree just before pickling. Sphinx events are queued so each extension runs it's own `process_doc` before Sphinx starts writing the doctree out.

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -135,10 +135,6 @@ def doctree_read(app, doctree):
             app.info(bold('[AutoAPI] ') +
                      darkgreen('Adding AutoAPI TOCTree [%s] to index.rst' % toc_entry)
                      )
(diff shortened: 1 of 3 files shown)


agjohnson:
I suppose I don't understand the original code or the fix very well. Is it possible to write a test case that would pass before and after this change? I'd like to ensure we're not removing something that has undocumented historical significance.

AWhetter:
Because we edit the toctree, I think the lines were added to ensure that the autoapi entry is correctly stored in the toctree, but the lines don't seem necessary. I've added a couple of tests to ensure that autoapi exists in the output toc.

AWhetter:
Closing in favour of #121

PR #116, opened 2017-08-31
Py hierarchy

### LINKED ISSUE #96: Introduce the notion of a hierarchy
Currently autoapi dumps every "file" as a flat list under the toc tree. It would make more organisational sense to put each entry into a heirarchy. For Python this could be done by module, for C# (and Java/C++) by namespace.

So what autoapi currently does as

```
my_package
my_package.my_module
my_package.my_other_module
```

would become

```
my_package
    |
    |- my_package.my_module
    L  my_package.my_other_module
```

AWhetter:
This is really two bits of work, that I can split up if wanted.

Firstly, Python documentation is now presented in a hierarchy. Only the top level package is displayed at the top level. Each subsequent subpackage and submodule is included in the toctree of the parent package.

Secondly, I've changed the templates to include a summary of the page, organised into the type of thing being documented (class, method, etc). As part of this, I've added an `autoapisummary` summary directive that does a similar thing to `autosummary`, but hooks into autoapi instead of loading the modules. Usage of the `autoapisummary` directive in templates requires the autosummary extension to be loaded in Sphinx, because we produce an autoaummary table node in the doctree.

Fixes #96


AWhetter:
Rebased on to top of #121 to fix the tests and make this easier to review. Only the last 4 commits are different from #121.

ericholscher:
Getting this on output:

```
Exception occurred:
  File "/Users/eric/.virtualenvs/pydoc/lib/python3.5/site-packages/sphinx/writers/html.py", line 869, in unknown_visit
    raise NotImplementedError('Unknown node: ' + node.__class__.__name__)
NotImplementedError: Unknown node: autosummary_table
```

ericholscher:
Ah, because it requires autosummary now. Not sure the best practice for that, and makes it hard to install I think.

ericholscher:
Can we import & register the node instead of requiring users to include the module? That seems much better.

AWhetter:
We can! We can call `app.setup_extension('sphinx.ext.autosummary')` in `setup()` of the extension. I don't know if it is usable with languages other than Python. Should I get it to load only if we're using Python or always?

ericholscher:
Just a heads up, I have a combined merge of all this stuff in this branch: https://github.com/rtfd/sphinx-autoapi/compare/combined-merge and I'm testing that locally :)

ericholscher:
Good question, I just did it unconditionally for now.

autoapi/directives.py
```diff
@@ -1,9 +1,133 @@
...
 from docutils import nodes
 
+from sphinx import addnodes
+import sphinx.ext.autosummary
 from sphinx.util.nodes import nested_parse_with_titles
+try:
+    from sphinx.util.rst import escape
+except ImportError:
+    def escape(text):
+        return re.compile(r'([!-/:-@\[-`{-~])').sub(r'\\\1', text)
```
agjohnson: This could use a comment on reason for implementation, for our future selves.

docs/conf.py
```diff
@@ -20,7 +20,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['autoapi.extension']
+extensions = ['autoapi.extension', 'sphinx.ext.autosummary']
```
agjohnson: +1 on moving the addition of this to our extension. We shouldn't force users to have to specify this extension to use autoapi.

autoapi/templates/python/module.rst
```diff
@@ -10,10 +10,65 @@
...
+{% block functions %}{% if obj.functions %}
+Functions
+~~~~~~~~~
+.. autoapisummary::
+
+{% for function in obj.functions %}
+   {{ function.id }}
+{%- endfor %}
+{% endif %}{% endblock %}
+
```
agjohnson: If we make autoapisummary optional, templates will need to change to be conditional as well.

agjohnson:
Unblocked. Feel free to rebase

PR #118, opened 2017-09-01
Added basic class signature discovery

AWhetter:
Implements a part of #99

### DIFF
--- a/autoapi/mappers/python.py
+++ b/autoapi/mappers/python.py
@@ -85,7 +85,7 @@ class PythonPythonMapper(PythonMapperBase):
 
         # Optional
         self.children = []
-        self.args = []
+        self._args = []
         if self.is_callable:
             self.args = self._get_arguments(obj)
         self.docstring = obj.docstring
@@ -96,6 +96,14 @@ class PythonPythonMapper(PythonMapperBase):
 
         # For later
         self.item_map = defaultdict(list)
+
+    @property
+    def args(self):
+        return self._args
+
+    @args.setter
+    def args(self, value):
+        self._args = value
 
     @property
     def is_undoc_member(self):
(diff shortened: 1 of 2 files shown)


AWhetter:
Rebased on to top of #121 to fix the tests and make this easier to review.

autoapi/mappers/python.py
```diff
@@ -267,6 +275,17 @@ class PythonPackage(PythonPythonMapper):
...
     type = 'class'
 
+    @PythonPythonMapper.args.getter
+    def args(self):
+        if self._args:
+            return self._args
+
+        for child in self.children:
+            if child.short_name == '__init__':
+                return child.args
```
agjohnson: If this is only for the case where this object is a class type, perhaps this should be more explicit about the fact, instead of just looking for `__init__`.

autoapi/templates/python/class.rst
```diff
@@ -1,4 +1,4 @@
-.. py:class:: {{ obj.short_name }}{% if obj.args %}({{ obj.args|join(',') }}){% endif %}
+.. py:class:: {{ obj.short_name }}{% if obj.args %}({{ obj.args[1:]|join(',') }}){% endif %}
```
agjohnson: Is this to remove `self` from the arguments?

autoapi/mappers/python.py
AWhetter: I don't understand what you mean.

autoapi/templates/python/class.rst
AWhetter: Yep!

autoapi/mappers/python.py
agjohnson: The case above is implying that any object with a child `__init__` can be called with the arguments of the `__init__`.

What it should say is if the object type is a class, and has a child `__init__`, the object can be called with arguments of the `__init__`.

AWhetter: This property override exists on the `PythonClass` mapper so it applies to classes only. `PythonPythonMapper` uses `self._args` only.

agjohnson: Roger. This is plenty then

agjohnson:
Unblocked. Will need a rebase

agjohnson:
Thanks for the rebase! LGTM

PR #119, opened 2017-09-12
Fixed usage of deprecated status_indicator

AWhetter:
`app.status_indicator` is deprecated in Sphinx 1.6, and is removed for the upcoming 1.7 version. This switches to using `sphinx.utils.status_indicator` instead as recommended by Sphinx (http://www.sphinx-doc.org/en/stable/changes.html#id9).


AWhetter:
Good point. It's not so I've added a version check.

AWhetter:
Closing in favour of #121

PR #121, opened 2017-11-05
Fix tests

AWhetter:
With the tests failing by default, this fixes them all so that pull requests are easier to review. This supersedes #114  and #119, and includes the fixes for both.
This also adds tests for Python 3.6 and sphinx 1.6.

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -135,10 +135,6 @@ def doctree_read(app, doctree):
             app.info(bold('[AutoAPI] ') +
                      darkgreen('Adding AutoAPI TOCTree [%s] to index.rst' % toc_entry)
                      )
-            if sphinx.version_info >= (1, 5):
-                app.env.toctree.process_doc(app.env.docname, doctree)
-            else:
-                app.env.build_toc_from(app.env.docname, doctree)
 
 
 def setup(app):
--- a/autoapi/mappers/dotnet.py
+++ b/autoapi/mappers/dotnet.py
@@ -60,6 +60,7 @@ class DotNetSphinxMapper(SphinxMapperBase):
 
     DOCFX_OUTPUT_PATH = '_api'
 
+    # pylint: disable=arguments-differ
     def load(self, patterns, dirs, ignore=None, **kwargs):
         '''Load objects from the filesystem into the ``paths`` dictionary.
 
@@ -124,7 +125,7 @@ class DotNetSphinxMapper(SphinxMapperBase):
         return None
 
     # Subclassed to iterate over items
-    def map(self, options=None, **kwargs):
+    def map(self, options=None):
         '''Trigger find of serialized sources and build objects'''
         for path, data in self.paths.items():
             references = data.get('references', [])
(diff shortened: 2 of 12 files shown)


AWhetter:
Looks like AppVeyor is failing with an unrelated error! :(

autoapi/extension.py
```diff
@@ -135,10 +135,6 @@ def doctree_read(app, doctree):
             app.info(bold('[AutoAPI] ') +
                      darkgreen('Adding AutoAPI TOCTree [%s] to index.rst' % toc_entry)
                      )
-            if sphinx.version_info >= (1, 5):
```
ericholscher: I don't believe we can just delete this..

autoapi/mappers/dotnet.py
```diff
@@ -98,8 +97,6 @@ def load(self, patterns, dirs, ignore=None, **kwargs):
                     self.app.warn(error_output)
             except (OSError, subprocess.CalledProcessError) as e:
                 self.app.warn('Error generating metadata: {0}'.format(e))
-                if raise_error:
-                    raise ExtensionError('Failure in docfx while generating AutoAPI output.')
```
ericholscher: Similarly, I don't believe we can just remove this logic, any reason for it?

autoapi/extension.py
AWhetter: There was some explanation in #114 but I should have copied that over to here.
> Because we edit the toctree, I think the lines were added to ensure that the autoapi entry is correctly stored in the toctree, but the lines don't seem necessary. I've added a couple of tests to ensure that autoapi exists in the output toc.

I added the tests to test_integration.py.

autoapi/mappers/dotnet.py
AWhetter: I wasn't really sure what to do with this. I removed it because the linter was complaining that the addition of `**kwargs` meant that the function signature didn't match the base class. Ideally we want some way of setting `raise_error` through the base class rather than for dotnet specifically.
Perhaps we add a Werror/strict setting where warnings are raised as errors?

agjohnson: What is the lint error on `**kwargs`? Seems we should keep this logic and work around the lint error instead.

tests/test_domains.py
```diff
@@ -86,7 +86,7 @@ def _mock_read(self, path):
         with patch('autoapi.mappers.dotnet.DotNetSphinxMapper.find_files', _mock_find):
             with patch('autoapi.mappers.dotnet.DotNetSphinxMapper.read_file', _mock_read):
                 dom = dotnet.DotNetSphinxMapper(self.application)
-                dom.load('', '', '', raise_error=False)
```
agjohnson: More `raise_error` removal to potentially address.

autoapi/extension.py
agjohnson: I feel okay with this change with the tests validating the toc 👍

autoapi/mappers/dotnet.py
AWhetter: It complains that the method signature is different from the method on the base class, because we've added a new argument(s). I felt that adding a proper way of configuring whether to raise an error made more sense than adding kwargs. The extension interacts with the mappers in the same way for each language, so to me it makes more sense for the base class to be providing this type of configuration.

agjohnson: Hrm, yeah not sure the best way to get around this. For now, feel free to revert, ignore the lint error, and open up an issue with some background so we can get back to addressing this more properly.

ericholscher:
:+1:

Issue #126, opened 2017-11-06
javascript domain: unrecognized jsdoc output structure


fredbi:
Using jsdoc v3.5.5 from npm package.

Produces a concluding record for JS file which is not recognized. Got warning.

```
[AutoAPI] Mapping Data
WARNING: Unknown Type: {u'files': [u'D:\\src\\js\\app.js'], u'kind': u'package', u'longname': u'package:undefined'}
WARNING: Unknown Type: {u'files': [u'D:\\src\\js\\example.js'], u'kind': u'package', u'longname': u'package:undefined'}
WARNING: Unknown Type: {u'files': [u'D:\\src\\js\\fundsdkapi.js'], u'kind': u'package', u'longname': u'package:undefined'}
[AutoAPI] Rendering Data
```

```
    {
        "kind": "package",
        "longname": "package:undefined",
        "files": [
            "D:\\src\\js\\app.js"
        ]
    }
```

Filtering it mutes the warning. I have no idea of the impact on doc, though.

In mappers/javascript.py, the following filter cleans the output.

```
for item in (i for i in data if not u'files' in i):
```

agjohnson:
Thanks for the note on the fix! I'm guessing this is an issue with the version of jsdoc we are expecting output from? If you feel like putting together a PR for this, we should be able to support multiple versions of jsdoc output.

fredbi:
I am not sure of the result. Filtering makes it work, but in my opinion it would be better to support a new sphinx macro for package files.
I prepare a PR with the quick fix (filter out annoying fields) since it is frustrating for new users.

AWhetter:
Closing this now that we are dropping support for documenting languages other than Python. See #248 for more information on this decision.

PR #130, opened 2017-11-13
126 unrecognized jsdoc tag

fredbi:
PR for issue #126.
This is a quick and dirty fix to avoid frustration when using jsdoc versions producing a list of files for a package ("file:"). This attribute is not interpreted yet however, just filtered out.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -144,6 +144,47 @@ any file matching ``['project.json', 'csproj', 'vsproj']`` will be searched for.
 .. _`ASP.NET Installation`: http://docs.asp.net/en/latest/getting-started/index.html
 .. _`docfx configuration file`: https://dotnet.github.io/docfx/tutorial/docfx.exe_user_manual.html#3-docfx-json-format
 
+
+GO
+~~~~
+
+Install go domain extension for sphinx.
+
+.. code:: bash
+  pip install go sphinxcontrib_golangdomain
+
+
+Install the go environment (from https://golang.org/dl/).
+
+Install a git client for your environment (e.g. from https://desktop.github.com/).
+
+Install our godocjson tool (preprocess godoc output to JSON, in a way similar to jsdoc -X).
+
+.. code:: bash
+  go get github.com/rtfd/godocjson
+
+Add go domain in your conf.py.
+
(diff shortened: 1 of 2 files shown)


fredbi:
This is a lint issue related to code previously here.

autoapi/mappers/javascript.py
```diff
@@ -35,7 +35,7 @@ def read_file(self, path, **kwargs):
     def map(self, options=None):
         '''Trigger find of serialized sources and build objects'''
         for path, data in self.paths.items():
-            for item in data:
+            for item in (i for i in data if not u'files' in i):
```
ericholscher: The lint error is about this, it should be: `for item in (i for i in data if u'files' not in i):`

fredbi: I follow your recommendation but have not way of checking this.

autoapi/mappers/javascript.py
```diff
@@ -35,7 +35,7 @@ def read_file(self, path, **kwargs):
     def map(self, options=None):
         '''Trigger find of serialized sources and build objects'''
         for path, data in self.paths.items():
-            for item in data:
+            for item in (i for i in data if u'files' not in i):
```
AWhetter: Please could you add a comment here for why the check is included.
Also, I'm being picky but, please could you move the `if` to inside the for loop. Looping over a generator defined on the loop itself is a bit unconventional.

README.rst
```diff
@@ -144,6 +144,47 @@ any file matching ``['project.json', 'csproj', 'vsproj']`` will be searched for.
...
+Install nodejs on your platform.
+
+Install jsdoc using npm.
+
+.. code:: bash
+  npm install jsdoc -g
+
+Before building your doc, make sure the jsdoc executable is in your path.
+
+
```
AWhetter: Looks like this could do with being rebased? These changes already exist in master.

AWhetter:
I'm going to close this because the original author cannot update the pull request, and we have no way of testing the changes.
However someone else with the original issue can feel free to pick this up and make a fix themselves.

PR #131, opened 2017-11-14
Extended ast parsing with astroid

### LINKED ISSUE #99: Expand ast parsing to allow for inheritance docs, data docs, and class signatures

AWhetter:
This implements extended ast parsing with astroid to attempt to resolve #99. On its own this adds:
* Inheritance documentation, including resolution of the full name from imports. So a class `class MyClass(Base):` would show `Bases: package.module.Base`.
* Discovery of data documentation (ie module level constants and class attributes), including attempting to display the value of the attribute like autodoc does.


AWhetter:
Are there any thoughts on this. I've put this here more to start a discussion than to get it merged right away.

AWhetter:
All of the tests are now passing for this, and this is no longer a proof of concept but a complete working change.
I've started using this on a few projects. The only quirk that I've noticed so far is that the source directories need to exist in the Python path so that astroid can successfully convert the file path of a module to it's proper module name. This is a requirement when using relative imports in the source code.

ericholscher:
Amazing. I will try and find some time to review this in the next week or so!

AWhetter:
Please could this be looked at soon. It will become more difficult to maintain the more it diverges.

ericholscher:
Tried this out locally, and it looks good. It would be good to be able to configure some of this stuff (eg. disable `Bases` being output or similar.)

Thanks for your work!

Issue #108, opened 2017-04-08
Prefix objects with module name


agjohnson:
The objects don't seem to have the top level module name as part of the object name. This is confusing UX

AWhetter:
Closing as fixed in #131

PR #135, opened 2018-04-21
Support specifying package directories in `autoapi_dirs`

slai:
At the moment, if you build the docs in this repo, the autoapi docs are rooted by their subpackages, i.e. listed as `mappers` instead of `autoapi.mappers`. The latter is the one that's expected, as when you import, you do `import autoapi.mappers`. The incorrect module path permeates through all the docs.

This change fixes that by changing the `autoapi_dirs` config parameter - if a directory specified has a `__init__.py` under it, the directory itself will also be included in the relative paths, resulting in `autoapi.mappers`. The previous behaviour is still there though, as if the directory does **not** have a `__init__.py` under it, the children will be documented with their path as the root, i.e. `mappers`.

I've added tests to demonstrate this.

I believe this addresses #108. Please let me know if the approach isn't right - I may have missed something that already allows this


AWhetter:
Another change has been merged that fixes the problem that you are describing. Please could you try it and let us know if it works for you.
It also works by looking for `__init__.py` files, but this will probably have to change in a future release to support [implicit namespace packages](https://www.python.org/dev/peps/pep-0420/) in Python 3.

slai:
@AWhetter It doesn't work. I think this is because parts of the mapper are still overwriting the name with a name based on the `relative_path` field, which my PR changes, but yours do not, e.g.
https://github.com/rtfd/sphinx-autoapi/blob/77c71b6cb73f75decb60039216287835c55d0ed6/autoapi/mappers/python.py#L254-L263
and 
https://github.com/rtfd/sphinx-autoapi/blob/77c71b6cb73f75decb60039216287835c55d0ed6/autoapi/mappers/python.py#L241-L248

Changing the `_resolve_name` functions to do nothing results in a blank document set. I haven't had time to investigate why, but I guess `name` is possibly used for path resolution too later on?

**All this said**, I've discovered that I can workaround this by reaching nto the original parser obj to get the correct name, e.g. changing the template from 

```
.. py:module:: {{ obj.name }}
```

to 

```
.. py:module:: {{ obj.obj.name }}
```

(Note the extra `.obj` part.)

I'll see if I can find out what's happening this weekend, although you may have some ideas too. Thanks for looking at this!

autoapi/mappers/python.py
```diff
@@ -240,7 +244,7 @@ class PythonModule(TopLevelPythonPythonMapper):
 
     def _resolve_name(self):
         name = self.obj['relative_path']
-        name = name.replace('/', '.')
+        name = name.replace(os.sep, '.')
```
slai: This is needed to make it work on Windows - `relative_path` can contain multiple components and without this, the replacement doesn't occur and causes tests to fail.

AWhetter:
Ah I think I understand your problem now. Are you pointing autoapi_dirs directly at the package directory? I think this will be a necessity to support implicit namespace packages so I'm happy for this to be the normal behaviour, as long as we document it as such.
I'll review with this in mind but if you want me to merge with the behaviour that you have here than let me know and I can make the changes later.

autoapi/mappers/python.py
```diff
@@ -36,10 +36,14 @@ def load(self, patterns, dirs, ignore=None):
         shortened, relative path the package/module
         """
         for dir_ in dirs:
+            dir_root = dir_
+            if os.path.exists(os.path.join(dir_, '__init__.py')):
```
AWhetter: We can remove this `if` and make the following line the default behaviour.
Please could you add a test for a package with an implicit namespace as well.

tests/pypackageexample/Makefile
```diff
@@ -0,0 +1,195 @@
+# Makefile for Sphinx documentation
```
AWhetter: This file can be removed. We don't need it for the test.

tests/test_integration.py
```diff
@@ -125,6 +125,42 @@ def test_integration(self):
                 index_file
             )
 
+    def test_integration_with_package(self):
```
AWhetter: This is testing a lot of similar things to what exists already. Could we remove this redundancy and check that the full names are correct instead? Perhaps including a module level function as well?

It may make sense to put the existing pyexample test into a test class with a setup and teardown instead of building two sets of docs?

docs/config.rst
```diff
@@ -10,6 +10,9 @@ Configuration Options
 
         Paths (relative or absolute) to the source code that you wish to generate your API documentation from.
```
AWhetter: Can we reword this section to make it clearer that we expect the package directory to be specified? Maybe mention implicit namespace packages as well.

slai:
> Are you pointing autoapi_dirs directly at the package directory?

Doh. Yes, but given this hint, I've just tried pointing autoapi_dirs at the directory _containing_ the package directory, and it worked as desired without the changes here. I did have to exclude a few files, e.g. setup.py.

Let me have a read up on implicit namespace packages and I'll work out what's needed here before making further changes.

AWhetter:
@slai Were you still thinking of implementing support for implicit namespace packages?

slai:
@AWhetter sorry for the delay; I'll take a look on Monday. Most likely I'll just address the comments you mentioned and we can look at implicit namespace packages in a separate change.

slai:
@AWhetter let me know what you think of the changes. If catering for single module distributions aren't something we want to worry about for now, then I can change this to support implicit namespace packages too.

AWhetter:
Looks great. Thanks for the change!

PR #136, opened 2018-04-21
Fix lint errors

slai:
There should be no behavioural changes - this is just explicitly returning what would've been returned anyway.


davidfischer:
Thank you for these fixes!

PR #137, opened 2018-05-23
Added support for viewcode

AWhetter:
This relies on Sphinx 1.8 which is yet to be released: http://www.sphinx-doc.org/en/master/ext/viewcode.html#event-viewcode-find-source

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -13,12 +13,18 @@ from sphinx.addnodes import toctree
 from sphinx.errors import ExtensionError
 from docutils.parsers.rst import directives
 
+from . import utils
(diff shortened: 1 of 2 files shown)


autoapi/extension.py
```diff
@@ -151,12 +157,45 @@ def clear_env(app, env):
...
+    return result
+
+
 def setup(app):
     app.connect('builder-inited', run_autoapi)
     app.connect('doctree-read', doctree_read)
     app.connect('doctree-resolved', add_domain_to_toctree)
     app.connect('build-finished', build_finished)
     app.connect('env-updated', clear_env)
+    if 'viewcode-find-source' in app.events.events:
```
ericholscher: Could use a comment here -- I'm guessing this is the guard against `<1.8`?

PR #138, opened 2018-06-06
Can control how __init__ docstring is displayed

### LINKED ISSUE #107: Merge __init__ into class definition
Currently, these are listed separately, so there is a class definition and then an `__init__` object under that listing. The two should be merged under the class definition.

AWhetter:
This adds a new `autoapi_python_class_content` option that works similarly to the [`autoclass_content`](http://www.sphinx-doc.org/en/master/ext/autodoc.html#confval-autoclass_content) option from autodoc.

Closes #107

### DIFF
--- a/autoapi/mappers/python.py
+++ b/autoapi/mappers/python.py
@@ -83,8 +83,13 @@ class PythonSphinxMapper(SphinxMapperBase):
         except KeyError:
             self.app.warn("Unknown type: %s" % data['type'])
         else:
-            obj = cls(data, jinja_env=self.jinja_env,
-                      options=self.app.config.autoapi_options, **kwargs)
+            obj = cls(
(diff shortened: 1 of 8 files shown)


autoapi/mappers/python.py
```diff
@@ -146,6 +152,14 @@ def args(self):
     def args(self, value):
         self._args = value
 
+    @property
+    def docstring(self):
+        return self._docstring
+
+    @docstring.setter
```
ericholscher: Where does this come from?

AWhetter: `self.docstring` is set in the constructor already. I've had to turn this into a property so that `PythonClass` can customise how this is accessed and use the `__init__` docstring if configured.

PR #149, opened 2018-08-02
Autodoc directives

### LINKED ISSUE #111: Add autodoc style directives for autoapi usage
Using directives for this will allow for manual inclusion of auto api docs. This should still allow for a full path to be created, perhaps as a secondary option once we have this feature.

### LINKED ISSUE #148: Fix compatibility with astroid 2.0
The Python 3 tests are failing since starting to use astroid 2: https://travis-ci.org/rtfd/sphinx-autoapi/jobs/410098167

AWhetter:
For now I have implemented these as autodoc Documenters because the autodoc framework gives us a lot for free including the option flags, traversal and filtering of members, and rendering.

If we later decide that we want these directives to following templating like autoapi does by default then we can add `auto<type>.rst` templates. I don't really know how useful templates would be. Wither they would generate only the header of the object (eg `.. py:class: MyClass(blah)\n   :noindex:`), or they'll have to contain the logic behind option usage and which members get shown but this seems annoying to work with. The flexibility of the directives comes from the options and laying out the directives yourself, not so much from auto-generation.

This includes some other fixes as well including:
* Fixing the tests under Python 2
* Fixing compatibility with astroid 2 under Python 3 (Closes #148)
* Can resolve the full name of a base class for a class defined in the same module
* Including `self` in method arguments (autodoc does so it seemed to make sense for us to do the same)
* Static and class methods are documented as such using their relevant directive

Closes #111


autoapi/documenters.py
```diff
@@ -0,0 +1,220 @@
...
+            return obj.docstring
+
+        for child in obj.children:
+            if child.name == name:
+                return child
+
+        if defargs:
+            return defargs[0]
+
+        raise AttributeError(name)
```
ericholscher: Does this get hit often? I wonder if it should have a bit more of a log message in order for users to debug this.

PR #161, opened 2019-02-20
Properly pass  file patterns to godocjson utility

orn0t:
Passing **autoapi_ignore** to godocjson utility exclude parameter while parsing with GoSphinxMapper.


AWhetter:
Nice and simple. Thanks for the change!

PR #165, opened 2019-06-18
Fix an issue with function whitespace causing weirdness

ericholscher:
You can see this live here until it's fixed: https://sphinx-notfound-page.readthedocs.io/en/latest/autoapi/notfound/extension/index.html#notfound.extension.html_collect_pages

This is what it currently looks like:

Screenshot 2019-06-18 13 01 10

It outputs RST that looks like this:

```
.. function:: replace_uris(app, doctree, nodetype, nodeattr)
   Replace ``nodetype`` URIs from ``doctree`` to the proper one.

   Rest of the docstring
```

Where we really want the first space there.

I tried a couple other fixes, but this one was the only one that seemed to work reliably.

### DIFF
--- a/autoapi/templates/python/function.rst
+++ b/autoapi/templates/python/function.rst
@@ -1,6 +1,8 @@
 {% if obj.display %}
 .. function:: {{ obj.short_name }}({{ obj.args }}){% if obj.return_annotation is not none %} -> {{ obj.return_annotation }}{% endif %}
 
+   {# Force Jinja to honor a newline here #}
+   {{''}} 
    {% if obj.docstring %}
    {{ obj.docstring|prepare_docstring|indent(3) }}
    {% endif %}


autoapi/templates/python/function.rst
```diff
@@ -1,6 +1,8 @@
 {% if obj.display %}
 .. function:: {{ obj.short_name }}({{ obj.args }}){% if obj.return_annotation is not none %} -> {{ obj.return_annotation }}{% endif %}
 
+   {# Force Jinja to honor a newline here #}
+   {{''}}
```
AWhetter: I am being picky but with what's here I'm getting some trailing whitespace in the generated rst. Ideally the rst output is as human friendly as possible so this might bother some users.
I've used a double empty line in the other template files to get the spacing to look like we want. Some quick testing looks like that should work here too?

setup.py
```diff
@@ -19,7 +19,7 @@
         'astroid<2;python_version<"3"',
         'Jinja2',
         'PyYAML',
-        'sphinx>=1.6',
+        'sphinx>=1.6,<2',
```
AWhetter: Was something broken with Sphinx 2 for you? The tests are passing with Sphinx 2 and I'm using it locally without issues.

autoapi/templates/python/function.rst
ericholscher: Feel free to use another approach, I just couldn't get them working locally.

setup.py
ericholscher: Yea, tests in travis were failing with them. I realize that probably shouldn't be in this PR, but I just wanted to see if this change was breaking tests. I can remove it 👍

AWhetter:
I've merged this manually and I am releasing it as part of v1.1.0. Thanks for the fix!

ericholscher:
Great, thanks!

PR #171, opened 2019-09-01
WIP: Fixes #170

### LINKED ISSUE #170: Make Go and .NET dependencies optional
Hello,

Currently **sphinx-autoapi** hardly depends on *sphinxcontrib-golangdomain* and *sphinxcontrib-dotnetdomain* libraries. As example, i use this library only for documenting Python codebase and i not willing to install any extra packages. Mostly because this packages or their dependencies may conflict with other project dependencies. And it's become harder when you are using poetry and have no options to deal with this potential situation.

From negative side, this change will introduce backward-incompatible changes, all everyone, who generates documentation for Go or .NET and have plain `sphinx-autoapi` or `sphinx-autoapi==^1.1.0` in their requirements, will lose ability to generate docs after next update.

danields761:
As i described in #170, it's really annoying that this library hardly depends on sphinxcontrib-golangdomain and sphinxcontrib-dotnetdomain, so i made a fix attempt and leave it at your discretion

Changes:

* Makes sphinxcontrib-golangdomain and sphinxcontrib-dotnetdomain to be optional packages
* Adds check that appropriate dependencies of a specified api-type installed during the extension initialisation

### DIFF
--- a/autoapi/backends.py
+++ b/autoapi/backends.py
@@ -34,26 +34,4 @@ backend_requirements = {
     "javascript": (),
     "go": (("sphinxcontrib-golangdomain", "sphinxcontrib.golangdomain"),),
     "dotnet": (("sphinxcontrib-dotnetdomain", "sphinxcontrib.dotnetdomain"),),
-}  # type: (Dict[str, Sequence[Tuple[str, str]]])
-
-
-def _get_available_backends():
-    def is_importable(name):
-        try:
-            __import__(name)
-            return True
-        except ModuleNotFoundError:
-            return False
-
-    backends = set()
-    for backend_name, deps in backend_requirements.items():
-        if all(is_importable(import_name) for _, import_name in deps):
-            backends.add(backend_name)
-
-    return backends
-
-
-available_backends = _get_available_backends()
-
-
-del _get_available_backends
+}  # type: Dict[str, Sequence[Tuple[str, str]]]
--- a/autoapi/backends.py
+++ b/autoapi/backends.py
@@ -25,3 +25,35 @@ default_backend_mapping = {
     "go": GoSphinxMapper,
     "javascript": JavaScriptSphinxMapper,
 }
+
+
+#: describes backend requirements in form
+#: {'backend name': (('1st package name in pypi', '1st package import name'), ...)}
+backend_requirements = {
+    "python": (),
+    "javascript": (),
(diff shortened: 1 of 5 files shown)


autoapi/backends.py
```diff
@@ -25,3 +25,35 @@
...
+#: {'backend name': (('1st package name in pypi', '1st package import name'), ...)}
+backend_requirements = {
+    "python": (),
+    "javascript": (),
+    "go": (("sphinxcontrib-golangdomain", "sphinxcontrib.golangdomain"),),
+    "dotnet": (("sphinxcontrib-dotnetdomain", "sphinxcontrib.dotnetdomain"),),
+}  # type: (Dict[str, Sequence[Tuple[str, str]]])
+
+
+def _get_available_backends():
```
AWhetter: Do you think it will simplify this if we check for the module in `sys.modules` instead (probably in `autoapi/extension.py` rather than here)? A user should have specified the domain extension in the `extensions` list in their `conf.py` already, so we can check that the extension is loaded.

autoapi/extension.py
```diff
@@ -67,6 +69,23 @@ def run_autoapi(app):
     )
     url_root = os.path.join("/", app.config.autoapi_root)
 
+    # check backend availability
+    # in reality if user will attempt to use `go` api type without
+    #  sphinxcontrib.golangdomain installed, sphinx will prohibit
+    #  all operations because of sphinxcontrib.golangdomain
+    #  extension is required
```
AWhetter: I don't think this comment is needed. You have made the error message nice and clear, so it is obvious what the code is checking without needing a comment.

autoapi/backends.py
danields761: That's is good point, will do in this way

danields761:
@AWhetter, thank for your replies

I fixed raised remarks. Also i added "autoapi_type" setting check, which ensures this value is valid, as well as test-cases covering changes (using pytest, if you don't mind).

P.s. i see failed pipline, but don't see were i was wrong :)

AWhetter:
The linter is broken at the moment so don't worry about it.
Thanks for your change!

danields761:
Thank you for merging this changes!

I think that this change should be covered inside documentation: optional dependencies available, ex: `pip install sphinx-autoapi[go,dotnet]`. But i don't think i can write it clear.

Issue #168, opened 2019-07-04
Problem running the Go example with sphinx v2.1.2


ArmandBENETEAU:
Dear all,

We would like to use your sphinx-autoapi tool to auto-document our Go code. It would be very useful for our company since we would use the Sphinx to document our python code and our Go code.

I have Sphinx v2.1.2 running on my Ubuntu 18.04.2 LTS laptop, and it works really well to document python 3.6 code. 

Therefore, I tried to creating the documentation of your Go example code by doing the following procedure : 

1/ Getting your example from your GIT repository

2/ Creating the following repository structure and getting the Sphinx Makefile : 
.
├── build
├── Makefile
└── source
........├── conf.py
........├── example
..................└── godocjson.go
........├── index.rst
........├── _static
........└── _templates

Sphinx Makefile : 

``` make
# Minimal makefile for Sphinx documentation
#

# You can set these variables from the command line, and also
# from the environment for the first two.
SPHINXOPTS    ?=
SPHINXBUILD   ?= sphinx-build
SOURCEDIR     = source
BUILDDIR      = build

# Put it first so that "make" without argument is like "make help".
help:
	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

.PHONY: help Makefile

# Catch-all target: route all unknown targets to Sphinx using the new
# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
%: Makefile
	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

Unfortunately, when I run the command `make html`, a lot of warnings and an exception are appearing. This result in no documentation being created. You can find here below the complete logs. I would like to know what is happening, is it possible to resolve it and if yes, how I can resolve this exception?

``` bash
Sphinx v2.1.2 en cours d'exécution
création du répertoire de sortie... fait
[AutoAPI] Loading Data
[AutoAPI] Mapping Data
[AutoAPI] Rendering Data
construction en cours [mo]:targets for 0 po files that are out of date
construction en cours[html]: targets for 1 source files that are out of date
updating environment: 3 added, 0 changed, 0 removed
lecture des sources...[100%] index                                                                                                                                                                                 
/home/armand/Bureau/Test/source/autoapi/main/index.rst:21: WARNING: Explicit markup ends without a blank line; unexpected unindent.
[...]

PR #172, opened 2019-09-03
Correction on the "Go" documentation creation with "sphinx-autoapi"

ArmandBENETEAU:
These corrections allow to run the example of Go code documentation with the latest version of Sphinx.

Nevertheless, it appears that if we want to use python 3 to launch the documentation generation, the repository "sphinxcontrib-golangdomain" needs to have correction as well. Indeed, some of the code in this directory is python 2 exclusive (iteritems() especially).

This pull request follow the Issue #168


AWhetter:
This is amazing! Thank you for the fix!

Issue #173, opened 2019-09-04
Filtering entities


danields761:
Hello,

I'm interesting in a feature similar to *sphinx.ext.autodoc* [autodoc-skip-member](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-skip-member) which gives precise control over entities included in result document.

PR #174, opened 2019-09-04
WIP: Implements experimental arbitrary members ignoring capability

danields761:
As i said in #173, feature to ignore members based on some specific conditions is very desirable. There is prototype of such feature involving *sphinx* events mechanism, work only with Python. I achieved that by emiting `autoapi-skip-member` event which must return something evaluable to `True` for excluding current entity. Event handler signature aims to be as close as possible to the appropriate [autodoc-skip-member](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-skip-member) handler.

Pease, take a look :)

### DIFF
--- a/autoapi/mappers/python/objects.py
+++ b/autoapi/mappers/python/objects.py
@@ -1,10 +1,12 @@
 import collections
-import logging
+from typing import Optional
 
-import sphinx
-from lazy_object_proxy.utils import cached_property
+import sphinx.util.logging
 
 from ..base import PythonMapperBase
+
+
+LOGGER = sphinx.util.logging.getLogger(__name__)
 
 
 class PythonPythonMapper(PythonMapperBase):
@@ -21,11 +23,12 @@ class PythonPythonMapper(PythonMapperBase):
     language = "python"
     is_callable = False
 
-    def __init__(self, obj, class_content="class", **kwargs):
+    def __init__(self, obj, parent=None, class_content="class", **kwargs):
         super(PythonPythonMapper, self).__init__(obj, **kwargs)
 
         self.name = obj["name"]
         self.id = obj.get("full_name", self.name)
+        self.parent = parent  # type: PythonPythonMapper
 
         # Optional
         self.children = []
@@ -35,6 +38,8 @@ class PythonPythonMapper(PythonMapperBase):
 
         # For later
         self._class_content = class_content
+
+        self._display_cache = None  # type: Optional[bool]
 
     @property
     def args(self):
@@ -93,7 +98,7 @@ class PythonPythonMapper(PythonMapperBase):
         """
         return self.short_name.startswith("__") and self.short_name.endswith("__")
 
-    @cached_property
+    @property
     def display(self):
         """Whether this object should be displayed in documentation.
 
@@ -102,16 +107,11 @@ class PythonPythonMapper(PythonMapperBase):
 
         :type: bool
         """
-        skip_undoc_member = self.is_undoc_member and "undoc-members" not in self.options
-        skip_private_member = (
-            self.is_private_member and "private-members" not in self.options
-        )
-        skip_special_member = (
-            self.is_special_member and "special-members" not in self.options
-        )
+        if self._display_cache is not None:
+            return self._display_cache
 
-        skip = skip_undoc_member or skip_private_member or skip_special_member
-        return not self._ask_ignore(skip)
+        result = self._display_cache = not self._ask_ignore(self._should_skip())
+        return result
 
     @property
     def summary(self):
@@ -129,13 +129,24 @@ class PythonPythonMapper(PythonMapperBase):
 
         return ""
 
+    def _should_skip(self):  # type: () -> bool
+        skip_undoc_member = self.is_undoc_member and "undoc-members" not in self.options
+        skip_private_member = (
+            self.is_private_member and "private-members" not in self.options
+        )
+        skip_special_member = (
+            self.is_special_member and "special-members" not in self.options
+        )
+
+        return skip_undoc_member or skip_private_member or skip_special_member
+
     def _ask_ignore(self, skip):  # type: (bool) -> bool
         try:
             ask_result = self.app.emit_firstresult(
                 "autoapi-skip-member", self.type, self.id, self, skip, self.options
             )
         except Exception as exc:
-            logging.error(
+            LOGGER.error(
                 'Exception occurs while evaluating "autoapi-skip-member" '
                 'event hook for "{}"'.format(self),
                 exc_info=exc,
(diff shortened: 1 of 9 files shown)


autoapi/mappers/python/objects.py
```diff
@@ -100,6 +100,22 @@ def display(self):
...
         :type: bool
         """
+        try:
+            skip = (
+                self.app.emit_firstresult(
+                    "autoapi-skip-member",
+                    self._class_content,
+                    self.name,
+                    self,
+                    self.options,
```
AWhetter: Should we move this until after the following `if` statements so that we can pass their result as a `skip` option to the event?

Something like

```python
skip = False
if self.is_undoc_member and ...:
    skip = True
elif self.is_private_member and ...:
    skip = True
elif ...:
    ...

skip = self.app.emit_firstresult(..., skip, self.options)
return not skip
```

autoapi/mappers/python/objects.py
```diff
@@ -183,10 +199,7 @@ def display(self):
 
         :type: bool
         """
-        if self.short_name == "__init__":
-            return False
-
-        return super(PythonMethod, self).display
+        return super(PythonMethod, self).display and self.short_name != "__init__"
```
AWhetter: This might need to fire it's own version of the event if we want to be able to also pass the result of `self.short_name != "__init__"` into the `skip` argument of the event?

autoapi/mappers/python/objects.py
```diff
@@ -100,6 +100,22 @@ def display(self):
 
         :type: bool
         """
+        try:
+            skip = (
+                self.app.emit_firstresult(
+                    "autoapi-skip-member",
+                    self._class_content,
```
AWhetter: Should this be `self.type`? `self._class_content` stores the value of the `autoapi_python_class_content` option rather than the type of object. You might need to make sure that `type` is correct for each type of object.

autoapi/mappers/python/objects.py
```diff
@@ -100,6 +100,22 @@ def display(self):
...
+                    self.name,
```
AWhetter: `self.full_name` stores the fully qualified name rather than `self.name`.

autoapi/mappers/python/objects.py
danields761: Ok, i will write appropriate tests

autoapi/mappers/python/objects.py
danields761: That is good point :)

autoapi/mappers/python/objects.py
danields761: There is no such attribute `self.full_name` on `PythonMapperBase` instance, maybe you mean `self.id` attribute?

AWhetter: Woops. Yes I do mean `self.id`!

autoapi/mappers/python/objects.py
danields761: I not really sure about this, generally i don't understand why this rule is hard-wired, previously i thought that it depends on `autoapi_python_class_content` option. Then i realized, how it works :)
In my opinion, class docstring and `__init__` docstring should be represented as different objects, what you think about?

danields761:
@AWhetter, i fixed skip-member hooks signature. There is still some pending questions, i am waiting for their clarification :)

Also there is necessary to complete documentation about this changes and i need some help with it because of may English %)

autoapi/mappers/python/objects.py
AWhetter: Usually `__init__` isn't displayed because the arguments of `__init__` are displayed in the signature of the class itself, and the arguments are documented in the class docstring. Sometimes this isn't desirable, which is why `autoapi_python_class_content` exists.
But now that that you're adding this new event, users will have the option to display `__init__` anyway if they want.
So I think it makes sense to hide `__init__` by default, but still ask your new event if it should be displayed.

autoapi/mappers/python/objects.py
```diff
@@ -1,6 +1,8 @@
 import collections
+import logging
 
 import sphinx
+from lazy_object_proxy.utils import cached_property
```
AWhetter: Please can we remove this dependency. `display` shouldn't be getting accessed very often anyway, but even if it is we need to investigate it separately.

autoapi/mappers/python/objects.py
```diff
@@ -1,6 +1,8 @@
 import collections
+import logging
```
AWhetter: autoapi uses the sphinx logging module so that we get the log formatting provided by sphinx. You can look at some of the other modules to see how we use it.

autoapi/mappers/python/objects.py
danields761: I got you, but still wandering: why you still want to merge class docstring with neither `__init__` or `__new__` methods docstrings? I think that depending on `autoapi_python_class_content` options they should be displayed independently. As example:

*class* **Foo**
&nbsp;&nbsp; Class docstring

&nbsp;&nbsp; **\_\_init\_\_(**_self, arg: bool_**)**
&nbsp;&nbsp;&nbsp;&nbsp; Additional doctring as well

Then it will be pretty obvious how to class methods :)

autoapi/mappers/python/objects.py
danields761: As i saw in my tests, for some members `display` method called two or more times. Not very much, to be honest. I did it to prevent confusing end-user when his callback called many times for same members. As examples, user may want to count something on per-member basis and such behaviour will be surprising. As option i may cache `display` property result manually.

autoapi/mappers/python/objects.py
AWhetter: Usually you don't use `__init__` directly (or `__new__`). You use them by the class itself. For example you use `instance = MyClass()` instead of `instance = MyClass.__init__()`. So it doesn't make sense to document it as a method.

autoapi/mappers/python/objects.py
```diff
@@ -299,13 +329,8 @@ def args(self):
...
 
-        if self._class_content in ("both", "init"):
-            constructor_docstring = self.constructor_docstring
-            if constructor_docstring:
-                if self._class_content == "both":
-                    docstring = "{0}\n{1}".format(docstring, constructor_docstring)
-                else:
-                    docstring = constructor_docstring
+        if self._class_content not in ("class", "both"):
+            return ""
```
AWhetter: Why did this need to change?

autoapi/mappers/python/objects.py
```diff
@@ -174,19 +205,18 @@ def __init__(self, obj, **kwargs):
...
-            return False
+        assert isinstance(
+            self.parent, PythonClass
+        ), "Method parent object is always class"
 
-        return super(PythonMethod, self).display
+        if self._class_content == "class":
+            return True
+        else:
+            return self.name == "__new__" and self.parent.constructor is not None
```
AWhetter: I think this whole function can simply be

```python
skip = self.name in ("__new__", "__init__")
return self._ask_ignore(skip)
```

autoapi/mappers/python/objects.py
```diff
@@ -329,6 +354,13 @@ def constructor(self):
 
         return None
 
+    @property
+    def new_constructor(self):
```
AWhetter: It looks like this isn't used anywhere?

AWhetter:
I've merged this in a114af76dc7113f8d3f8e23ab36ebdcdce6814f0
Thanks for all of your work!

danields761:
@AWhetter, sorry i come too late!

Anyway, thanks for merging base functionality :)

PR #176, opened 2019-09-25
Go mapper improve

SilverRainZ:
- d10207d depends on https://github.com/readthedocs/godocjson/pull/13.
- 7ebb0c6 fixes the probelms that go method are not included in the generated rst, but when render rst to html I get `WARNING: Unknown directive type "go:method"`, any way to fix it?

### DIFF
--- a/autoapi/mappers/go.py
+++ b/autoapi/mappers/go.py
@@ -128,6 +128,7 @@ class GoPythonMapper(PythonMapperBase):
             obj.get("parameters", []),
         )
         self.parameters = list(temp_parameters)
+        self.results = obj.get("results", [])
         self.docstring = obj.get("doc", "")
 
         # Go Specific
--- a/autoapi/mappers/go.py
+++ b/autoapi/mappers/go.py
@@ -172,6 +172,10 @@ class GoMethod(GoPythonMapper):
     type = "method"
     ref_directive = "meth"
 
+    def __init__(self, obj, **kwargs):
+        super(GoMethod, self).__init__(obj, **kwargs)
+        self.receiver = obj.get("recv")
+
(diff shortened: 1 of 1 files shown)


SilverRainZ:
BTW, the current rst tempate is toooooo simple.

AWhetter:
You can fix the unknown directive problem by installing `sphinxcontrib-golangdomain`, which gets installed when you install autoapi with `pip install sphinx-autoapi[go]`.
The documentation hasn't been updated to reflect this change yet.

SilverRainZ:
@AWhetter I already have sphinxcontrib-golangdomain installed, directive like `go:const` works well but `go:method` is unrecognized.

AWhetter:
The golangdomain docs say that methods are documented with `.. go:function: (type)`?

SilverRainZ:
Yes, but rst template use ref_type as directive name (`.. go:{{ obj.ref_type }}:: {{ obj.name }}`).

 And https://github.com/readthedocs/sphinx-autoapi/blob/master/autoapi/mappers/go.py#L172 use `method` as ref_type of go method, should I fix it?

AWhetter:
If you can then that would be great! :D

SilverRainZ:
It seems there is no a simple way to fix it, my suggestion is refactor the go template.
Before that, can you please review this PR and other 2 of godocjson?

SilverRainZ:
Any response?

AWhetter:
Sorry I haven't had time to look at this yet. We can't merge this until someone looks at the pull request for godocjson though. I don't have permissions to maintain that.

DerThorsten:
any progress here?

PR #179, opened 2019-10-23
Fixed table in the tutorial

Cerberus1746:
I realized that there was a `::` before the table, which screwed it up.


AWhetter:
Thanks for the fix! I've merged this as 2af3d46

Issue #213, opened 2020-06-04
scoped missing from template module.rst


lhfriedman:
In module.rst template, ``content`` is an outer block with inner blocks like ``classes``. The ``scoped``
keyword is missing so that the inner block processes empty lists. See: 

https://jinja.palletsprojects.com/en/2.11.x/templates/#block-nesting-and-scope_

I'm going to make a branch and fix this, unless any developers object.

PR #214, opened 2020-06-04
added scoped keywords

lhfriedman:
see issue 213 (https://github.com/readthedocs/sphinx-autoapi/issues/213)


lhfriedman:
Suggest that the ``:nosignature:`` option be added to the .. autoapisummary:: but that is tangential, not a bug.

AWhetter:
Good spot. Thanks for the fix!

PR #220, opened 2020-06-12
(tox) black diff color

MatthijsBurgh:


MatthijsBurgh:
`--color` seems not be released yet.

PR #221, opened 2020-06-13
Fix ci

MatthijsBurgh:
Don't find in class summary


AWhetter:
Nice find. Thanks for the fix!

PR #231, opened 2020-07-22
Added default golang ingore pattern to remove unit tests by default

DerThorsten:
With this, golang test files are ignored.
since godocjson is shall be called like this to ignore tests we do not need no wildcard symbol (*)
```
godocjson  -e _test.go fobarpath
```


AWhetter:
Thanks for the contribution!

Issue #227, opened 2020-07-12
Methods fail to get documented with autodoc-style directives in Sphinx 3.1+


AWhetter:
`tests/python/test_pyintegration.py::TestSimpleModule::test_manual_directives` fails and raises warnings such as

```
error while formatting signature for example.Bar.method_okay: 'NoneType' object has no attribute '__dict__'
```

It looks like support for overloads in autodoc has broken this test, so we probably need to implement our own overload support.

PR #235, opened 2020-07-31
Set self.parent in import_object

nuchi-airbnb:
Fixes https://github.com/readthedocs/sphinx-autoapi/issues/227

At least as far as I can tell. The error appears to originate in https://github.com/sphinx-doc/sphinx/blob/9a2f2e15a6c2149a84d91f34f18695086dc47f90/sphinx/ext/autodoc/__init__.py#L1771


nuchi-airbnb:
Thanks @AWhetter — I noticed that commented-out test only after opening the PR. I ran the test locally and unfortunately it didn't pass. I think I possibly might have fixed a different unrelated issue that was presenting with the same symptom — my use case now works but I don't have a clean minimum working example nor a good explanation for why. Sorry about that!

nuchi-airbnb:
Found my error — I needed to set `self.parent` only on methods.

PR #236, opened 2020-08-01
Fix #234: Inject TOCs just after doctrees are read

### LINKED ISSUE #234: Missing functions of package after `index`.
Hi hi! We're using AutoAPI in CircuitPython and are hitting a weird bug where functions aren't listed in a toctree consistently. I think this has to do with the order the files are processed.

Here is a small-ish repo case. Click into the `ulab` entry and you'll see that only half of the modules have functions listed underneath even though they have functions.

Screenshot_2020-07-27 ulab — Python documentation

[sphinx-bug.zip](https://github.com/readthedocs/sphinx-autoapi/files/4985473/sphinx-bug.zip)

Thanks!

ciscorn:
This PR moves the TOC-injection process from the writing phase to the resolving phase.

I think this will solve the issue discussed in #234.

(However, I'm afraid it causes any side-effects I haven't noticed. Are there any reasons to do this at `doctree-resolved`?)

Here is an example build result. Things after "index" are shown:

sc

Close #234

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -186,6 +186,9 @@ def doctree_read(app, doctree):
     """
     Inject AutoAPI into the TOC Tree dynamically.
     """
+
+    add_domain_to_toctree(app, doctree, app.env.docname)
+
     if app.env.docname == "index":
         all_docs = set()
         insert = True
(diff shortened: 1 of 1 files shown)


AWhetter:
You're totally right that this is the right place to do this. Nice find!

PR #245, opened 2020-09-08
Fix inheritance diagram from module

brentyi:
`astroid.Module` has a `get_children()` function, but no `children` field. (I think the `body` field should also work here)

http://pylint.pycqa.org/projects/astroid/en/latest/api/astroid.nodes.html#astroid.nodes.Module.get_children


AWhetter (approved):
Thanks for the fix!

PR #268, opened 2021-02-15
Added autoapi-skip-member example

j0yu:
Just a quick example to illustrate how to register events in Sphinx, which very subtly [mentioned in their official documentations](https://www.sphinx-doc.org/en/master/extdev/appapi.html?highlight=conf#sphinx-core-events) and may catch out new users.

image


AWhetter:
This was merged as aeecc3e3e49a9906a8d9a29c84c1bcf12468845f

PR #281, opened 2021-04-08
Added attribute summary table to python module template

### LINKED ISSUE #280: Group module attributes/constants in separate summary table
Currently documented module constants/attributes are listed with their values, but there is no summary table displayed. It would be great if a summary table could be added to these.

I believe this is because these constants/attributes exist in `visible_children` [here](https://raw.githubusercontent.com/readthedocs/sphinx-autoapi/master/autoapi/templates/python/module.rst), but there is no associated summary table generated by the template. I couldn't work out exactly how module attributes are identified and classified under the hood, but if they are, then it seems like one could simply define `visible_attributes` by filtering from `visible_children` and then adding the following section to the template:

```
{% block functions scoped %}
{% if visible_attributes %}
Attributes
~~~~~~~~~

.. autoapisummary::

{% for attribute in visible_attributes %}
   {{ attribute.id }}
{% endfor %}


{% endif %}
{% endblock %}
```

If that diagnosis and solution is correct, then I could take a pass at implementing this and submit it via a PR. But since I am unfamiliar with what's going on under the hood here I am unsure if this would actually work.

P.s. great package btw. Thanks for all the hard work 👌 👌 👌

sammorley-short:
Aims to fix #280. Tox passing (only 3.6 and 3.7, I didn't have 3.8 and 3.9)

However, it's a bit of a naive hack, so I suspect there's some additional stuff that needs doing here. So lemme know what else needs doing for this.

### DIFF
--- a/autoapi/templates/python/module.rst
+++ b/autoapi/templates/python/module.rst
@@ -60,6 +60,7 @@ Submodules
 
 {% set visible_classes = visible_children|selectattr("type", "equalto", "class")|list %}
 {% set visible_functions = visible_children|selectattr("type", "equalto", "function")|list %}
+{% set visible_attributes = visible_children|selectattr("type", "equalto", "data")|list %}
 {% if "show-module-summary" in autoapi_options and (visible_classes or visible_functions) %}
 {% block classes scoped %}
 {% if visible_classes %}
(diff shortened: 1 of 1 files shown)


AWhetter (approved):
This looks perfect to me. Thanks for the changes!

PR #284, opened 2021-04-16
Add option autoapi_add_objects_to_toctree

jakobandersen:
Add an option such that the automatic addition of objects to the toctree can be disabled by the user.

It's a nice feature, but is in a sense orthogonal to the goal of extracting documentation from source code.


AWhetter (changes requested):
Thanks for making this change. Please could we add some documentation to the "config" page for this new configuration option.

It would also be useful to know why you want to be able to turn off this behaviour.

AWhetter (approved):

PR #290, opened 2021-06-01
Fix class args formatting when no __init__

jakobandersen:
When documenting a class which has no ``__init__`` method there seems to be a missing initialization, resulting in the warning ``'list' object has no attribute 'startswith'``.
Can be reproduced with [autoapiTest.tar.gz](https://github.com/readthedocs/sphinx-autoapi/files/6580359/autoapiTest.tar.gz).


AWhetter:
Thanks for trying this out. I've made the proper fix for this problem in 5faec730737a4932143e0a34a37a081bae7166f5

jakobandersen:
Great, thanks!

PR #300, opened 2021-08-05
Fix comment typehints for methods/classmethods

mathbou:
When using comment type hints, `self` and `cls` parameter appears in method parameters and types do not match anymore.

Using the following code, here is the results:
```python
class Truc(object):
    def foo(self, arg1, arg2):
        # type: (str, int) -> None
        print(arg1, arg2)
```

With sphinx-autoapi 1.8.3:

With the fix:


fmigneault:
@AWhetter 
Is it possible to have feedback on this PR as potential fix for #318?
Thanks

AWhetter:
Hey @mathbou. Sorry it's taken me so long to look at this. The change looks pretty good to me, however there are some failing tests. Please could you fix those, and also rebase on top of master as there were some formatting changes made to conform to the new version of black being used.

mathbou:
@AWhetter Changes done, is it good for you ?

PR #311, opened 2021-10-30
A set fixed signature length in summaries

### LINKED ISSUE #278: Module summary signature max width?
It would be nice if there was a max character length for the signatures shown in the module summary, so that when they are long you get a `...)` like with Sphinx autosummary. Currently the summary becomes less helpful if one signature is long because you can't see the descriptions without scrolling right.

zmoon:
Use the Sphinx autosummary machinery ([`mangle_signature`](https://github.com/sphinx-doc/sphinx/blob/f42d8f7df6b41d257ca9583a9113df26d04b3c76/sphinx/ext/autosummary/__init__.py#L481)) to shorten the signatures, like done in [`Autosummary.get_items`](https://github.com/sphinx-doc/sphinx/blob/f42d8f7df6b41d257ca9583a9113df26d04b3c76/sphinx/ext/autosummary/__init__.py#L329).

Fixes #278

### DIFF
--- a/autoapi/directives.py
+++ b/autoapi/directives.py
@@ -3,7 +3,7 @@
 from docutils.parsers.rst import Directive
 from docutils import nodes
 
-from sphinx.ext.autosummary import Autosummary
+from sphinx.ext.autosummary import Autosummary, mangle_signature
 from sphinx.util.nodes import nested_parse_with_titles
 
 from .mappers.python.objects import PythonFunction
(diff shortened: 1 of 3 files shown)


zmoon:
I guess this might only work for Python though?

AWhetter (changes requested):
Hey @zmoon. Apologies for taking so long to look at this. This change looks great! Please could you add a test for it. Somewhere in the `tests/python/pyexample` fixture that gets tested by `tests/python/test_pyintegration.py` would be fine.

Issue #294, opened 2021-06-22
`source_suffix` fallback doesn't allow `dict`


mgeier:
This assumes that `source_suffix` is a `list`:

https://github.com/readthedocs/sphinx-autoapi/blob/613eff53a7801d94eb864f2e27924d7d5592556f/autoapi/extension.py#L162

Quite a long time ago, Sphinx added support for using a `dict` instead, and now this is the recommended usage, see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix.

ain-soph:
So maybe a quick fix would be `list(source_suffix)[0]` ?  
Or do any pre-checks to ensure the it won't be empty?

ain-soph:
Yes. It seems the repo authors don't take the case into consideration where `".rst" not in source_suffix.keys()`.

AWhetter:
This was already closed in #321. Proper markdown support will be added in #287

PR #321, opened 2022-01-18
Fix keyerror:  0

garfieldnate:
Setting `source_suffix = ".md"` led to the following cryptic error output:

    Handler <function run_autoapi at 0x10841fc10> for event 'builder-inited' threw an exception (exception: 0)

This issue was that `app.config.source_suffix` is an `OrderedDict` object, which cannot be subscripted
using `[0]`, etc. Instead, use `next(iter(...))` to get the first element of the dictionary.

### DIFF
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -159,7 +159,7 @@ def run_autoapi(app):  # pylint: disable=too-many-branches
         out_suffix = ".txt"
     else:
         # Fallback to first suffix listed
-        out_suffix = app.config.source_suffix[0]
+        out_suffix = next(iter(app.config.source_suffix))
 
     if sphinx_mapper_obj.load(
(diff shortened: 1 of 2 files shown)


garfieldnate:
Sorry there are not tests; if someone could provide some guidance on where best to add one, I'd be happy to follow up with additional changes.

AWhetter:
@garfieldnate Apologies for taking so long to look at this. The change looks great! Please could you add a test under `tests/python/test_pyintegration.py`. You can build the "pyexample" fixture (just like `TestSimpleModule.built` does) in a new test function and set `source_suffix` in the `confoverrides`.

garfieldnate:
Sorry, I wasn't able to build and run the project locally properly, but I've added a test and pushed a new commit.

AWhetter:
Just noting that this closes #294

PR #340, opened 2022-05-31
Fix "all args are kw-only"

### LINKED ISSUE #328: asterisk in first position in Python function signature is dropped
Python 3.9.10
sphinx-autoapi: 1.8.4
Sphinx: 4.4.0
astroid: 2.9.3

The asterisk keyword indicator in a Python function argument list is being dropped in the generated documentation, if it occurs before any other arguments.

module.py:
```py
def f(*, a, b):
    """f function"""
    pass

def g(a, *, b):
    """g function"""
    pass
```
generates the HTML below. Note the missing `(*, `.

Is this an `astroid` parsing problem? Should I/you file an issue there? Thanks.

Here are all the build files:
[sigs.zip](https://github.com/readthedocs/sphinx-autoapi/files/8110711/sigs.zip)

image

jepler:
Closes #328 

Also fixes a compatibility problem with Sphinx 5.0.0 'language = None'.


tannewt:
@ericholscher Is there something newer than autoapi that we should be using?

ericholscher:
@tannewt Not that solves the same problem of doing parse-only API doc generation. Unfortunately it's a hard problem, and one that we got initial grant funding for, but haven't been able to focus on in the long term.

AWhetter:
This has been merged as 0ac4dc5389f758d6d93460f76df20268cd6ba4c3

jepler:
Thank you!

PR #343, opened 2022-07-10
Mention data's types and values in correct place

PerchunPak:
So type hints can be correctly parsed and loaded.

For example, now intersphinx will create link to `str`.


AWhetter:
I've had to roll this back because it breaks the tests. Please can you fix those.

PR #362, opened 2022-12-23
Don't force member sort order to alphabetic

ashb:
sphinx.ext.autodoc has a `member-order:` directive, and this powers the `sort_member` function[[1]] to sort object members for us -- by removing the `sorted` call ourselves we allow the "bysource" mode to work: with the `sorted` call, since we don't set `self.analyzer` no extra sorting happens, so the `:member-order: bysource` setting has no effect.

The default sort mode for autodoc is alphabetical, so this should result in no changes to docs by default

[1]: https://github.com/sphinx-doc/sphinx/blob/984416247370a87ad947d89f334a8ce2d8ea6558/sphinx/ext/autodoc/__init__.py#L816-L836

### DIFF
--- a/autoapi/documenters.py
+++ b/autoapi/documenters.py
@@ -79,7 +79,7 @@ class AutoapiDocumenter(autodoc.Documenter):
         elif not self.options.inherited_members:
             children = (child for child in children if not child[1].inherited)
 
-        return False, sorted(children)
+        return False, children
 
 
 class _AutoapiDocstringSignatureMixin:  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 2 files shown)


ashb:
I have tested this manually like so -- and changing the `member-order` directive does change the output order.

```rst
.. autoapimodule:: airflow.policies
  :no-members:
  :members: task_policy, dag_policy, task_instance_mutation_hook, pod_mutation_hook
  :member-order: bysource
```

ashb:
I should add tests for this.

ashb:
@AWhetter cc if you have time to take a look that would be ace.

AWhetter:
Looks great! Thanks for the change.

AWhetter:
@ashb Sorry it took so long to get this released. There were some test failures that had to be addressed first. This just got released in v2.0.1. Let me know if you encounter any issues.

PR #384, opened 2023-05-18
Catch all exceptions when reading files

felixvd:
Importing a file with this content:
```python
from typing import TypedDict
MyDict = TypedDict("MyDict", {'badField': List[], 'goodField': int})
```
Results in an extremely unhelpful error message that does not mention the offending file:
```
Extension error (autoapi.extension):
Handler <function run_autoapi at 0x7fbc8d1be3a0> for event 'builder-inited' threw an exception (exception: Parsing Python code failed:
invalid syntax (<unknown>, line 2))
```
This change fixes the behavior.

For the record, the `Exception` that was not caught was of type `astroid.exceptions.AstroidSyntaxError`.

### DIFF
--- a/autoapi/mappers/python/mapper.py
+++ b/autoapi/mappers/python/mapper.py
@@ -322,8 +322,8 @@ class PythonSphinxMapper(SphinxMapperBase):
             else:
                 parsed_data = Parser().parse_file(path)
             return parsed_data
-        except (IOError, TypeError, ImportError):
-            LOGGER.debug("Reason:", exc_info=True)
+        except Exception as e:
+            LOGGER.debug("Reason: Received Exception of type " + str(type(e)) + ": " + str(e), exc_info=True)
             LOGGER.warning(
                 "Unable to read file: {0}".format(path),
                 type="autoapi",


AWhetter:
The build is failing because there's invalid syntax in the file. If we don't raise an exception, then users might assume that everything has been documented correctly, when actually the file with the offending syntax error won't have been documented.
Why is catching all exceptions beneficial?

felixvd:
> If we don't raise an exception, then users might assume that everything has been documented correctly

I am not sure I can follow your argument. Not raising an exception is the current behavior when catching `IOError`, `TypeError`, `ImportError`. They all cause files not to be documented and only log a warning. If you say this is undesirable, the current code should be fixed, e.g. like this:

```diff
-        except (IOError, TypeError, ImportError):
+        except (IOError, TypeError, ImportError) as e:
            LOGGER.debug("Reason:", exc_info=True)
            LOGGER.warning(
                "Unable to read file: {0}".format(path),
                type="autoapi",
                subtype="not_readable",
            )
+            raise e
        return None
```

I can add the `raise e` line to this PR if you prefer.

> Why is catching all exceptions beneficial?

1. Any exception causes the file to be undocumented, so they should all be caught and reported.
2. Currently, the user is not told which file needs to be fixed. That is a significant problem.

Issue #180, opened 2019-10-24
[Python] `Unknown type: placeholder` error


alanhdu:
When setting up `sphinx-autoapi` for one of our internal repositories, I found a strange warning that `autoapi` would pull up:

```
WARNING: Unknown type: placeholder
```

Digging into the debugger a little bit, it turns out that this is from code like:
```
from A.B.C import (a, b, c)
from A.B.C.D import d

a1 = a
b1 = b
c1 = c

d = d   # this turns out to be a problem
```

where the debugger section shows that `d` has a `placeholder` type:

```python
[0]   /opt/anaconda/envs/ctrldev/bin/sphinx-build(10)<module>()
-> sys.exit(main())
[1]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/cmd/build.py(288)main()
-> return make_main(argv)
[2]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/cmd/build.py(195)make_main()
-> return make_mode.run_make_mode(argv[1:])
[3]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/cmd/make_mode.py(167)run_make_mode()
-> return make.run_generic_build(args[0])
[4]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/cmd/make_mode.py(155)run_generic_build()
-> return build_main(args + opts)
[5]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/cmd/build.py(275)build_main()
-> args.tags, args.verbosity, args.jobs, args.keep_going)
[6]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/application.py(269)__init__()
-> self._init_builder()
[7]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/application.py(330)_init_builder()
-> self.events.emit('builder-inited')
[8]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/sphinx/events.py(103)emit()
-> results.append(callback(self.app, *args))
[9]   /opt/anaconda/envs/ctrldev/lib/python3.6/site-packages/autoapi/extension.py(129)run_autoapi()
-> sphinx_mapper_obj.map(options=app.config.autoapi_options)
[...]

PR #394, opened 2023-06-13
fix: Unknown type: placeholder bug #180

rickstaa:
This PR adds a fix for the `Unknown type: placeholder` that was given in #180. See [this comment](https://github.com/readthedocs/sphinx-autoapi/issues/180#issuecomment-1589208233).

### DIFF
--- a/autoapi/mappers/python/mapper.py
+++ b/autoapi/mappers/python/mapper.py
@@ -331,7 +331,10 @@ class PythonSphinxMapper(SphinxMapperBase):
         modules = {}
         for module in self.paths.values():
             children = {child["name"]: child for child in module["children"]}
-            modules[module["name"]] = (module, children)
+            if module["name"] not in modules:
+                modules[module["name"]] = (module, children)
+            else:  # Update the module if it already exists
+                modules[module["name"]][1].update(children)
 
         resolved = set()
         for module_name in modules:


AWhetter:
Do you know what situation this bug is triggered in?
Would you be able to add a test for it?

rickstaa:
@AWhetter, I can add code in [this bug example package](https://github.com/rickstaa/unknown_type_placeholder_bug) to the tests folder and create a test based on that 👍🏻. 

However, please remember that this use case is not typical in regular Python packages but does come up when people use external builders that combine Python modules programmatically (e.g. [catkin](https://github.com/readthedocs/sphinx-autoapi/issues/180#issuecomment-1589208233)). I tried to see whether this bug comes up using a [native namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/). Still, it only is triggered when people deviate from the [recommendations](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages) and add a `__init__.py` at the top folder of the namespace subpackage (see https://github.com/rickstaa/unknown_type_placeholder_bug_ns) 🤔. I yet have to find a more standard python use case where this problem happens.

rickstaa:
@AWhetter Why this happens is quite easy to see by looking at [this bug example package](https://github.com/rickstaa/unknown_type_placeholder_bug). The current codebase only expects to encounter a module once. As a result, the code in 

https://github.com/readthedocs/sphinx-autoapi/blob/0ac1e113594908bff61c0a95baa6b1899c5ecdc3/autoapi/mappers/python/mapper.py#L333

overwrites all the entries that were mapped in the first pass.

AWhetter:
Closing due to lack of follow up. Please feel free to resubmit if you are able to add a test case.

Issue #405, opened 2023-09-22
`__init__.pyi` treated as module


laggykiller:
`docs/conf.py`

```python
...
autoapi_dirs = [os.path.abspath('../src')]
...
```

`src/__init__.py`
```python
'''sample-python-package'''
__version__ = '1.0.0'
```

`src/submodule_foo/__init__.pyi`
```python
class Foo:
    def bar(self, a = int) -> None:
        """
        Sample function
        """
        ...
```

The generated documentation list `submodule_foo/__init__` as a module `__init__`, but it is expected that `submodule_foo` should be listed as a module.

Note that adding `autoapi_file_patterns = ["*.pyi", "*.py"]` does not solve the problem.

The expected behaviour is shown if we rename `src/submodule_foo/__init__.pyi` to `src/submodule_foo/__init__.py` (Change extension from `pyi` to `py`)

Alternative workaround is to change `docs/conf.py`
```python
autoapi_dirs = [
    os.path.abspath('../src'),
    os.path.abspath('../src/submodule_foo')
    ]
autoapi_python_use_implicit_namespaces = True
```

The generated documentation would be correct without changing extension from `pyi` to `py`

Related issue: https://github.com/readthedocs/sphinx-autoapi/issues/243

---

Real life example using my project: https://github.com/laggykiller/apngasm-python/tree/stub-to-sphinx

To build the documentation:
```bash
git clone --recursive https://github.com/laggykiller/apngasm-python.git
cd ./apngasm-python
git checkout b7ef4f3ce1296461e5f89e4b205a56efaeb7d086
cd ./docs
python install -r ./requirements.txt
make html
```

This would generate incorrect documentation with `__init__` listed as module.

Workaround to build the correct documentation, we have to do this before building documentation:
```bash
# Working directory: apngasm-python
mv ./src-python/apngasm_python/_apngasm_python/__init__.pyi ./src-python/apngasm_python/_apngasm_python/__init__.py
```

This would generate correct documentation with `_apngasm_python` listed as module.

Alternatively, change `docs/conf.py`:
```python
autoapi_dirs = [
    os.path.abspath('../src-python/apngasm_python'),
    os.path.abspath('../src-python/apngasm_python/_apngasm_python')
    ]
autoapi_python_use_implicit_namespaces = True
```

This would also generate correct documentation with `_apngasm_python` listed as module, without renaming `__init__.pyi` to `__init__.py`.

---

[...]

PR #406, opened 2023-09-22
fix: Correct handling of `__init__.pyi`

laggykiller:
Solves https://github.com/readthedocs/sphinx-autoapi/issues/405

Might be a good idea to cleanup a bit though?

### DIFF
--- a/autoapi/directives.py
+++ b/autoapi/directives.py
@@ -41,7 +41,6 @@ class AutoapiSummary(Autosummary):
 
 
 class NestedParse(Directive):
-
     """Nested parsing to remove the first heading of included rST
 
     This is used to handle the case where we like to remove user supplied
--- a/autoapi/extension.py
+++ b/autoapi/extension.py
@@ -2,6 +2,7 @@
 
 This extension allows you to automagically generate API documentation from your project.
 """
+
 import io
 import os
 import shutil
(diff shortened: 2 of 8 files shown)


autoapi/mappers/python/parser.py
```diff
@@ -24,7 +24,7 @@ def _get_full_name(self, name):
     def _parse_file(self, file_path, condition):
         directory, filename = os.path.split(file_path)
         module_parts = []
-        if filename != "__init__.py":
+        if filename != "__init__.py" or filename != "__init__.pyi":
```
AWhetter, line 27: This will always be true. This is equivalent to `if not (filename == "__init__.py" and filename == "__init__.pyi"):`, and a file cannot equal both names. Is this supposed to be an `and` rather than an `or`?

AWhetter (changes requested):
Thanks for making this change. It looks promising.

Please could you also add a test case for this scenario that fails without your changes.
You'll also need to add a release notes fragment with towncrier. The README explains how to do this.

autoapi/mappers/python/parser.py
laggykiller: Yes, you are right. I have added a commit for this problem

laggykiller:
Done, please check!

laggykiller:
@AWhetter Seems like checks have failed, what went wrong?

AWhetter:
This got merged as 823c146b3a3b359778a2435c1b7e5410336c57f0

PR #423, opened 2024-03-03
Preserve strings inside Literal type annotations

GeeTransit:
Hello,

I really appreciate the work put into this extension. It's very easy to set up and to customize.

Currently, if one has a type annotation with [``Literal``](https://typing.readthedocs.io/en/latest/spec/literal.html#literal-types), the strings inside are treated as if they were forward references, rather than constant strings.

Given this function:
```py
def func(arg: Literal['one', 'two', 'three']) -> str:
    return arg
```

The generated documentation is this:
```py
func(arg: Literal[one, two, three]) → str
```

Rather than what would be expected:
```py
func(arg: Literal['one', 'two', 'three']) → str
```

I changed `astroid_utils._resolve_annotation` to check if the subscript value is ``Literal`` and if so, return `.as_string()` for constants such as strings and bytes. I also added a few tests and a news fragment.

### DIFF
--- a/autoapi/mappers/python/astroid_utils.py
+++ b/autoapi/mappers/python/astroid_utils.py
@@ -416,7 +416,20 @@ def _resolve_annotation(annotation):
         # astroid.Index was removed in astroid v3
         if hasattr(astroid, "Index") and isinstance(slice_node, astroid.Index):
             slice_node = slice_node.value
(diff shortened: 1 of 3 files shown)


AWhetter (approved):
Thank you so much for making this change!

AWhetter:
The test failure has been addressed in `main` already.

Issue #393, opened 2023-06-10
Replace usage of deprecated `traverse` attribute on docutils nodes


AWhetter:
Running the tests outputs the following warning:

```
tests/test_integration.py: 4 warnings
tests/python/test_pyintegration.py: 33 warnings
  /home/ashley/workspace/sphinx-autoapi/.tox/py311/lib/python3.11/site-packages/autoapi/extension.py:191: PendingDeprecationWarning: nodes.Node.traverse() is obsoleted by Node.findall().
    nodes = list(doctree.traverse(toctree))
```

We need to address this before a release of astroid removes the method.

PR #424, opened 2024-03-05
Replace usage of deprecated docutils.nodes.Node.traverse

GeeTransit:
Sphinx 6.1.0 (the lowest supported version) requires `docutils>=0.18`, which happens to be when `Node.findall` was introduced. This means we can use `Node.findall` without errors. Fixes GH-393 (which mistakenly mentioned astroid instead of docutils).


AWhetter:
The test failure has been addressed in `main` already.

PR #431, opened 2024-04-04
feat: Add warning for missing import

### LINKED ISSUE #419: sphinx-autoapi silently skips unresolved python elements
## Observed behavior

In an rst file with content like:

```rst
Autodoc Directives
==================

.. autoapimodule:: example
    :members:
    :noindex:

Non-existing code references
----------------------------

.. autoapifunction:: test_example.bar_test
```

Where `example` is an existing python module in `autoapi_dirs`, and `test_example.bar_test` is just something I fabulated, `sphinx-build -W -b html` will happily run through with no warnings.

The resulting HTML page has an empty section "Non-existing code references", i.e. the broken code import is silently ignored.

## Expected behavior

`sphinx-autoapi` emits a warning for every unresolvable code import.

## Steps to reproduce

I have added a self-contained example and the list of installed packages.

I tried with sphinx-autoapi 3.0.0 and sphinx-autoapi 2.0.0, and sphinx 7.2.6.

Would be happy to learn if this is a bug or intended behavior. As this is definitely unwanted in our project, I would like to learn how to add a warning for unresolveable imports.

[python-autoapi-example.zip](https://github.com/readthedocs/sphinx-autoapi/files/14209644/python-autoapi-example.zip)

arwedus:
This warning is shown by sphinx build now when the "fullname" in the autoapi directive cannot be found in the imports, e.g. `.. autoapiclass:: autoapi.mappers.python.objects.PythonPythonMapper` give:

```
reading sources... [100%] reference/templates
WARNING: Failed to import class 'autoapi.mappers.python.objects.PythonPythonMapper' [autoapi.import]
```

Closes #419

### DIFF
--- a/autoapi/documenters.py
+++ b/autoapi/documenters.py
@@ -1,5 +1,7 @@
 import re
 
+import sphinx.util.logging
+
 from sphinx.ext import autodoc
 
 from ._objects import (
@@ -11,6 +13,9 @@ from ._objects import (
     PythonAttribute,
     PythonException,
 )
+
+
+LOGGER = sphinx.util.logging.getLogger(__name__)
 
 
 class AutoapiDocumenter(autodoc.Documenter):
@@ -33,11 +38,18 @@ class AutoapiDocumenter(autodoc.Documenter):
 
         raise AttributeError(name)
 
-    def import_object(self):
+    def import_object(self) -> bool:
+        """Imports and sets the object to be documented.
+
+        The object is searched in the autoapi_objects dict based on the fullname attribute of the documenter.
+
+        Returns:
+            bool: True if the object was successfully imported and set, False otherwise.
+        """
         max_splits = self.fullname.count(".")
+        objects = self.env.autoapi_objects
         for num_splits in range(max_splits, -1, -1):
             path_stack = list(reversed(self.fullname.rsplit(".", num_splits)))
-            objects = self.env.autoapi_objects
             parent = None
             current = objects.get(path_stack.pop())
             while current and path_stack:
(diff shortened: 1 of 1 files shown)


autoapi/documenters.py
```diff
@@ -50,6 +62,9 @@ def import_object(self):
                 self._method_parent = parent
                 return True
 
+        # If we get here, the object was not found. Emit a warning as autodoc does.
+        LOGGER.warning("Failed to import %s '%s' [autoapi.import]", self.directivetype, self.fullname, type="autoapi", subtype="import")
+        self.env.note_reread()
```
AWhetter, line 67: What is this line for?

autoapi/documenters.py
```diff
@@ -50,6 +62,9 @@ def import_object(self):
                 self._method_parent = parent
                 return True
 
+        # If we get here, the object was not found. Emit a warning as autodoc does.
+        LOGGER.warning("Failed to import %s '%s' [autoapi.import]", self.directivetype, self.fullname, type="autoapi", subtype="import")
```
AWhetter, line 66: I believe that the inclusion of `[autoapi.import]` at the end of the message isn't required because the message type is output by default.

AWhetter (changes requested):
Apologies for taking so long to look at this.
In addition to my other comments, please could you add a test for this. You'll also need to use towncrier to add a changelog entry.

AWhetter:
Closing due to lack of follow up. Please feel free to resubmit if you can fix the tests and address the comments.

arwedus:
Hi @AWhetter , I'm picking this up again because we lost our "missing import" warnings again after an update to Autoapi 😄 

You mentioned a missing towncrier entry, but https://sphinx-autoapi.readthedocs.io/en/latest/maintenance/release-process.html does not mention towncrier anywhere. Can you be more specific?

I'll also add a test.

autoapi/documenters.py
arwedus: no idea how this came in. I'll delete it.

PR #432, opened 2024-04-04
fix: Correctly include templates API docs

arwedus:
My new warning implemented in https://github.com/readthedocs/sphinx-autoapi/pull/431 shows that the autoapi references in the templates.rst are actually broken.


AWhetter (approved):
Thanks for making this change. Please could you fix the tests by adding a changelog entry.

AWhetter:
I ended up integrating this as part of some other changes.

PR #433, opened 2024-04-04
Add test and docs extra dependencies to devcontainer

arwedus:
It's more convenient to have everything available in the devcontainer to build the autoapi docs or run tox for test directly after building the devcontainer.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,7 @@ docs =
     furo
     sphinx
     sphinx_design
+test = tox
 
 [options.packages.find]
 include =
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -44,6 +44,7 @@ docs =
     furo
     sphinx
     sphinx_design
+test = tox
```
AWhetter, line 47: Thanks for making this change. Rather than adding a new extra, please could you install tox directly in the `postCreateCommand`. Ultimately I want to get rid of the `docs` extra as well because I don't think that it's a good practise to put development requirements in extras.

AWhetter (changes requested):

AWhetter:
Closing due to lack of follow up. Please feel free to resubmit if you are able to address the comments above.

PR #455, opened 2024-06-14
Fix _static paths getting corrupted during RST generation

### LINKED ISSUE #454: html_static_path variable doesn't work as expected due to RST nesting.
I believe the `html_static_path` variable isn't working as expected on sphinx-autoapi. Sphinx copies the `_static` folders correctly into the build but the RST structure generated by autoapi is nested and paths aren't automatically adjusted to match that. This results in the `_static` paths needing to be modified for any file to be imported correctly.

For instance, the following example doesn't import the image correctly withou manual modifications: 

doc_package.cli
```python
def another_function(a: int, b: List[float]):
  """A function

  Args:
    a: An integer
    b: a list of floats

  Returns:
    Nothing

    .. image:: _static/img/an_image.png
      :scale: 50 %
  """
  pass

```

It generates the following RST:

autoapi/doc_package/cli/index.rst
```RST

.. py:function:: another_function(a, b)

   A function

   :param a: An integer
   :param b: a list of floats

   :returns: Nothing 

             .. image:: _static/img/an_image.png
               :scale: 50 %
```

Given that `_static` is several directories higher, for the image to import properly, it would need to be `../../../_static/img/an_image.png`.

sachahu1:
This fixes  #454

### DIFF
--- a/autoapi/_mapper.py
+++ b/autoapi/_mapper.py
@@ -4,6 +4,7 @@ import fnmatch
 import itertools
 import operator
 import os
+import pathlib
 import re
 
 from jinja2 import Environment, FileSystemLoader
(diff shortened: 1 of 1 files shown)


AWhetter:
Closing as per the comment in #454

PR #462, opened 2024-07-08
Mark tests that use the network

QuLogic:
These two tests uses intersphinx, and the marker allows skipping them if there is no network available.

### DIFF
--- a/docs/changes/462.misc.rst
+++ b/docs/changes/462.misc.rst
@@ -1,0 +1,1 @@
+Mark tests that use the network
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,11 @@ disallow_untyped_defs = true
 # Ignore until astroid is typed (https://github.com/pylint-dev/astroid/issues/2060)
 warn_return_any = false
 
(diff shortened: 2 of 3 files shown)


AWhetter:
Thanks for making this change. Why is this needed?

QuLogic:
This is useful for downstream packaging; in such cases builds aren't allowed to access the Internet at large, and instead of skipping all tests, it is useful to skip only the ones that require networking. I don't know that `network` is a standard name anywhere, but at least it's the one I suggest when making PRs like this.

cclauss:
To get the GitHub Actions to pass...
```bash
echo "Mark pytests that use the network" > docs/changes/462.misc.rst
git add docs/changes/462.misc.rst
git commit -m"git add docs/changes/462.misc.rst"
git push
```

QuLogic:
Thanks for the ping; I've fixed that, and the conflict.

AWhetter:
Thanks for fixing this yourself and keep the pull request up to date. I've merged this as 0d5f21b8650d6202208f0cf00f498a16a1e32379

PR #484, opened 2024-09-30
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Fixes software supply chain safety warnings by automating pull requests like https://github.com/readthedocs/sphinx-autoapi/commit/367cd96adad18ffe2208250dad1e1b00440074a0

* [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)
* [Configuration options for the dependabot.yml file - package-ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*"  # Group all Actions updates into a single larger pull request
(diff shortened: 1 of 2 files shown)


AWhetter:
I've always been put off from using this because of how noisy it is. Can we reduce the amount of pull requests that it creates? Maybe by using the groups option?
Also please could we use a monthly update interval.

cclauss:
The Groups option is already enabled in this PR, meaning that only one Dependabot PR can be open at a time—not chatty.

Dependabot would only be noisy if numerous GitHub Actions dependencies were pinned to minor or micro versions.

The current GitHub Actions dependencies are:
    - uses: actions/checkout@v4
    - uses: actions/setup-python@v5
    -  uses: actions/upload-artifact@v4
    -  uses: actions/download-artifact@v4
    -  uses: pypa/gh-action-pypi-publish@release/v1
    -  uses: sigstore/gh-action-sigstore-python@v3.0.0
 
sigstore/gh-action-sigstore-python is the only dependency pinned to a micro version.  The other most chatty would be Python which would have only been in four Dependabot PRs since GitHub Actions began.

PR #490, opened 2024-10-12
Fix ruff compressions and performance issues

cclauss:
% `ruff check --select=C4,PERF --statistics .`
```
3	C408   	[*] unnecessary-collection-call
2	PERF401	[ ] manual-list-comprehension
1	C400   	[*] unnecessary-generator-list
1	PERF203	[ ] try-except-in-loop
1	PERF403	[ ] manual-dict-comprehension
```
% `ruff rule PERF401`
# manual-list-comprehension (PERF401)

Derived from the **Perflint** linter.

## What it does
Checks for `for` loops that can be replaced by a list comprehension.

## Why is this bad?
When creating a transformed list from an existing list using a for-loop,
prefer a list comprehension. List comprehensions are more readable and
more performant.

Using the below as an example, the list comprehension is ~10% faster on
Python 3.11, and ~25% faster on Python 3.10.

Note that, as with all `perflint` rules, this is only intended as a
micro-optimization, and will have a negligible impact on performance in
most cases.

## Example
```python
original = list(range(10000))
filtered = []
for i in original:
    if i % 2:
        filtered.append(i)
```

Use instead:
```python
original = list(range(10000))
filtered = [x for x in original if x % 2]
```

If you're appending to an existing list, use the `extend` method instead:
```python
original = list(range(10000))
filtered.extend(x for x in original if x % 2)
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     strategy:
+      fail-fast: false
       matrix:
         # Keep this in sync with tox.ini
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,6 @@ on:
 jobs:
   test:
     strategy:
-      fail-fast: false
       matrix:
         # Keep this in sync with tox.ini
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
--- a/autoapi/inheritance_diagrams.py
+++ b/autoapi/inheritance_diagrams.py
@@ -42,11 +42,11 @@ def _import_class(name, currmodule):
         return [target]
 
     if isinstance(target, astroid.Module):
-        classes = []
-        for child in target.get_children():
-            if isinstance(child, astroid.ClassDef):
-                classes.append(child)
-        return classes
+        return [
+            child
+            for child in target.get_children()
+            if isinstance(child, astroid.ClassDef)
+        ]
 
     raise sphinx.ext.inheritance_diagram.InheritanceException(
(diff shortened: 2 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     strategy:
+      fail-fast: false
```
AWhetter, line 10: Please can we leave this on. It seems wasteful to run everything if something has failed.

autoapi/inheritance_diagrams.py
```diff
@@ -42,11 +42,11 @@ def _import_class(name, currmodule):
         return [target]
 
     if isinstance(target, astroid.Module):
-        classes = []
-        for child in target.get_children():
-            if isinstance(child, astroid.ClassDef):
-                classes.append(child)
-        return classes
+        return [
```
AWhetter, line 45: I'm not a fan of multi-line list comprehensions. I think we sacrifice readability for a minimal performance benefit. Please could this stay as-is.

docs/changes/490.misc.rst
```diff
@@ -0,0 +1 @@
+Fix ruff compressions and performance issues, limit new version of sphinx, add Py312,Py313 to trove classifiers
```
AWhetter, line 1: The sphinx issue has been addressed and the classifiers have been added since this was written. Please could you correct these.

AWhetter (changes requested):

.github/workflows/main.yml
cclauss: Wasteful of compute time or developer time? `fail-fast: false` saves developers time because they see all failing jobs in a single run and can understand the root cause faster and be more efficient in fixing.  If this helps them make fewer commits then it also reduces compute time.

`fail-fast: true` is a good default for GitHub because it reduces compute time however it forces the developer to go thru multiple commits when debugging multiple problems.,

Issue #312, opened 2021-11-11
Excluding directories


evildmp:
AutoAPI makes the reasonable assumption that one's directory structure will look like:

```
    mypackage/
    ├── docs directory
    │   ├── ...
    │   └── ...
    ├── code directory
    │   ├── ...
    │   ├── ...
    └── ...
```

I have at least one structure however that looks like:

```
    mypackage/
    ├── docs directory
    │   ├── ...
    │   └── ...
    ├── simple code file
    ├── another simple code file
    └── ...
```

Using ``autoapi_dirs = ['..']`` means that it descends back into ``docs`` where it also finds an ``env`` and gobbles up everything it finds in there. Then I get a ``Relative import with too many levels`` (and it spends a long time processing the contents of the env too).

Unfortunately ``autoapi_ignore`` only appears to work on filenames.

I'd like a way to be able to exclude all of some arbitrary directory, in my case, ``/docs``

rexwangcc:
Ran into a similar issue and I tried something like `*vendor/**/*` which did the trick for me, just FYI 😄

AWhetter:
To any first time contributors looking to resolve this issue, we can filter the list of directories here: https://github.com/readthedocs/sphinx-autoapi/blob/17ffa2d1b63bb36b8eb55e8fd22e6707a57e7a5b/autoapi/mappers/base.py#L239

rajkaramchedu:
This is a little beyond my coding skills, but I could really use a better way to exclude files and directories. The trouble with `*somekeyword*` is it will exclude all the files that match it, including the ones in the directory that I am including. 
**Use case:** If I am editing docstrings in a large Python library, I prefer the local builds go faster, and for that I need to exclude a lot of stuff.

PR #503, opened 2024-12-02
Allow skipping entire directories

bgmeulem:
See issue #312 

This feature proposal allows skipping entire directories using `autoapi_ignore`.
Instead of recursing into the directory and then flagging every single file for skipping, it simply does not recurse into the directory.
Tested on a large-scale codebase and works.

### Issue
I did encounter one weird edgecase during testing: faulty code is handled differently. I have one file with a handful of "X is not defined" errors that I haven't gotten around to fixing yet. `autoapi` did not used to complain about these and simply parsed out the docstrings.
With this change, I encountered 

```
Handler <function run_autoapi at 0x7fbae49edc10> for event 'builder-inited' threw an exception (exception: Inference failed for all members of [<Call l.1902 at 0x7fba90784cd0>].)
```
L1902 points to roughly the second object in the file that is not defined. Both the first and second non-defined objects are inits of different non-existing classes from within a class init. So not only does the error handling change when skipping entire directories, it also appears to be inconsistent (the first undefined object yields no issues). I suppose this error originates from `ast`? I sadly don't know enough about how `ast` works to figure out exactly what changes and why skipping directories rather than files in those directories would be different.

I'm also using `ast` for the `skip-member` hook, which may also be at the cause. Again, i don't know enough about the order of events during docs building to investigate this further. I am willing to try if someone can direct me a little bit more to potential causes.

### DIFF
--- a/autoapi/_mapper.py
+++ b/autoapi/_mapper.py
@@ -230,6 +230,21 @@ def _link_objs(value):
 
     # Strip off the extra "\ "
     return result[:-2]
+    
+def _path_matches_patterns(path, patterns):
+    """Check if a path matches one of multiple patterns
+
+    Args:
+        path (str): path to a file or directory to check
+        patterns (list): list of patterns for fnmatch
+
+    Returns:
+        bool: Whether or not the path matches a pattern in patterns
+    """
+    for pattern in patterns:
+        if fnmatch.fnmatch(path, pattern):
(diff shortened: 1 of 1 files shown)


AWhetter:
Sorry for taking so long to review this. This looks great! Thanks for making the change.

PR #522, opened 2025-05-06
A user tool to debug visibility of objects

robtaylor:
It seems users can get quite disoriented when switching off "undoc-members" - whole modules will stop rendering even though there may be documented objects within.

This PR adds a config that helps users to debug why items may not be rendering, and also adds an info log for the common case of the module itself missing top level documentation in __init__.py

Interested in your thoughts!

### DIFF
--- a/autoapi/_mapper.py
+++ b/autoapi/_mapper.py
@@ -30,6 +30,7 @@ from ._objects import (
     PythonData,
     PythonException,
     _trace_visibility,
+    HideReason,
 )
 from .settings import OWN_PAGE_LEVELS, TEMPLATE_DIR
 
@@ -45,12 +46,7 @@ LOGGER = sphinx.util.logging.getLogger(__name__)
 
 
 def _color_info(msg: str) -> None:
-    LOGGER.info(
-        colorize("bold", "[AutoAPI] ")
-        + colorize(
-            "darkgreen", msg
-        )
-    )
+    LOGGER.info(colorize("bold", "[AutoAPI] ") + colorize("darkgreen", msg))
 
 
 def _expand_wildcard_placeholder(original_module, originals_map, placeholder):
@@ -389,13 +385,6 @@ class Mapper:
         # Render Top Index
         top_level_index = os.path.join(self.dir_root, "index.rst")
 
-        modules = [obj for obj in self.all_objects.values()
-                   if obj.type == "module" and obj.docstring == ""]
-        if modules and "undoc-members" not in self.app.config.autoapi_options:
-            _color_info("The following modules have no top-level documentation, and so were skipped as undocumented:")
-            for m in modules:
-                _color_info(f"    {m.id}")
-
         pages = [obj for obj in self.objects_to_render.values() if obj.display]
         if not pages:
             msg = (
@@ -507,8 +496,12 @@ class Mapper:
                     and not obj["inherited_from"]["is_abstract"]
                     and module not in documented_modules
                 ):
-                    _trace_visibility(self.app, f"Hiding {obj['qual_name']} as determined to be Python standard Library (found as {obj['full_name']})", verbose=2)
-                    obj["hide"] = True
+                    _trace_visibility(
+                        self.app,
+                        f"Hiding {obj['qual_name']} as determined to be Python standard Library (found as {obj['full_name']})",
+                        verbose=2,
+                    )
+                    obj["hide_reason"] = HideReason.STD_LIBRARY
 
     def _resolve_placeholders(self):
         """Resolve objects that have been imported from elsewhere."""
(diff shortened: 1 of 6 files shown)


autoapi/_mapper.py
```diff
@@ -518,12 +527,18 @@ def _hide_yo_kids(self):
...
+                    _trace_visibility(self.app, f"    {n}")
                 for child in module["children"]:
                     if child["qual_name"] not in all_names:
+                        _trace_visibility(self.app, f"Hiding {child['full_name']}, as {child['qual_name']} not in __all__")
                         child["hide"] = True
             elif module["type"] == "module":
+                _trace_visibility(self.app, f"Testing if any children of {module['full_name']} have already been documented")
                 for child in module["children"]:
                     if "original_path" in child:
+                        _trace_visibility(self.app, f"Hiding {child['full_name']} as documented at {child['original_path']}")
```
AWhetter, line 541: These items are hidden not because they're documented elsewhere (in fact they may not be documented anywhere) because we don't consider objects imported from the local package into a module to be part of the public API. Usually those imports are used in the implementation of a class or function in that module instead.

autoapi/_objects.py
```diff
@@ -231,6 +243,22 @@ def _should_skip(self) -> bool:
             self.inherited and "inherited-members" not in self.options
         )
 
+        reason = ""
+        if self.obj.get("hide", False):
+            reason = "marked hidden by mapper"
```
AWhetter, line 248: This reason is too generic to be helpful. In addition, objects that are hidden this way will be logged twice. Once via a call to `_trace_visibility` where `hide` is originally set, and another in `display`.
Perhaps instead of only storing `obj["hide"] = True`, we should also store a `obj["hide_reason"]` for use by `display`?

AWhetter (changes requested):
This is a great idea! Thanks for making the change. I've left some comments for improvements.

autoapi/_objects.py
robtaylor: makes sense!

autoapi/extension.py
```diff
@@ -261,6 +261,7 @@ def setup(app):
     app.add_config_value("autoapi_generate_api_docs", True, "html")
     app.add_config_value("autoapi_prepare_jinja_env", None, "html")
     app.add_config_value("autoapi_own_page_level", "module", "html")
+    app.add_config_value("autoapi_verbose_visibility", 0, "html")
```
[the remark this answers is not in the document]
robtaylor: Actually, looking at this again, it adds way too much output to debug mode. I think it makes sense to have this separate for debugging visibility. Also I split the visibility debugging into two verbosity levels as usually level 1 will suffice, but allowing for deeper debugging,.

robtaylor:
I'm having some odd issues running the tests in `tox` - they're working when i run manually for 3.9 and 3.13, but failing under the `tox` runs. Any guidance appreciated!

autoapi/extension.py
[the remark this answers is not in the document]
AWhetter: In which case please can you add some documentation for this new config option to `docs/reference/config.rst`. You could also add a small section about debugging to `docs/how_to.rst`, possibly as a subsection to "How to Customise What Gets Documented", or linking to your new section from there.

AWhetter:
> One question I have - with the 'hide reason' enum idea, we could use this in autoapi-skip-member interface. Do you think it would make sense to add this to that api?

Good point. If we use an enum then someone can't extend that enum easily to add their own reasons for something being skipped. Perhaps letting `autoapi-skip-member` return a string (with the return type `bool | str | None`) would be best. What do you think?

robtaylor:
Hi @AWhetter, I've pushed two new commits addressing your feedback:

**1. Fixed the `display` cache logic** (`5f3c462`)
- The `else` branch now only traces when the object is actually hidden (not when `_display_cache` is `NOT_HIDDEN`)
- Uses `self._display_cache` instead of recomputing `self.hide_reason` on every access

**2. Added documentation** (`5f3c462`)
- Added `autoapi_verbose_visibility` config option docs to `docs/reference/config.rst` (under Advanced Options)
- Added a "How to Debug Visibility of Objects" subsection to `docs/how_to.rst`, under "How to Customise What Gets Documented"

**3. Documented the `__init__.py` docstring gotcha** (`17d7091`)
- Added a note under "Configure autoapi_options" explaining that packages without a module-level docstring in `__init__.py` are excluded when `undoc-members` is not enabled
- This covers the case I mentioned earlier about imports into `__init__` not being documented — the existing how_to section didn't mention this, so I think the note is a useful addition

Re your question about `autoapi-skip-member` returning `bool | str | None` — agreed, exposing as a string makes most sense since the enum is already a string enum. Happy to add that in a follow-up if you'd like.

PR #523, opened 2025-05-13
fix(typing): User may not have provided a return annotation for an overload

bzoracler:
If the source code doesn't provide a return type annotation on one or more overloads, then `1`-indexed value of the inner tuple, representing the annotation, can be `None`.

This is a consequence of the following:

https://github.com/readthedocs/sphinx-autoapi/blob/6dae5a3fa2d3c76cc28997a61d052376606a1bc1/autoapi/_parser.py#L459-L461

https://github.com/readthedocs/sphinx-autoapi/blob/6dae5a3fa2d3c76cc28997a61d052376606a1bc1/autoapi/_parser.py#L303

https://github.com/readthedocs/sphinx-autoapi/blob/6dae5a3fa2d3c76cc28997a61d052376606a1bc1/autoapi/_astroid_utils.py#L665

### DIFF
--- a/docs/changes/523.bugfix.rst
+++ b/docs/changes/523.bugfix.rst
@@ -1,0 +1,1 @@
+Fix `PythonFunction.overloads` typing when source code overload(s) do not provide a return type


AWhetter:
@bzoracler Thanks for making this change! Please could you fix the failing tests. It looks like you need to add release notes (see https://github.com/readthedocs/sphinx-autoapi?tab=readme-ov-file#release-notes).

PR #530, opened 2025-08-07
Render generic type params

jgrey4296:
This extracts `type_param` from the astroid parse, and uses it for generic classes, functions, and methods.
A `pytypeparams` simple sphinx project is added and used for testing.


AWhetter:
Thanks for this great contribution, and for waiting so patiently for me to review it!

Issue #536, opened 2025-10-05
Crash with astroid v4.0.0


bn-andrew:
A new release of astroid has just been released: https://github.com/pylint-dev/astroid/releases/tag/v4.0.0

Unfortunately they seem to have changed one of their APIs in a way that causes an error in autoapi.

```Traceback (most recent call last):                                                                                                                                                                                             
  File "/home/user/repos/repo/.venv/lib/python3.11/site-packages/autoapi/_mapper.py", line 475, in read_file                                                                                                   
    parsed_data = Parser().parse_file(path)                                                                                                                                                                                    
                  ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                    
  File "/home/user/repos/repo/.venv/lib/python3.11/site-packages/autoapi/_parser.py", line 44, in parse_file                                                                                                   
    return self._parse_file(                                                                                                                                                                                                   
           ^^^^^^^^^^^^^^^^^                                                                                                                                                                                                   
  File "/home/user/repos/repo/.venv/lib/python3.11/site-packages/autoapi/_parser.py", line 40, in _parse_file                                                                                                  
    node = astroid.builder.AstroidBuilder().file_build(file_path, module_name)                                                                                                                                                 
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                    
TypeError: AstroidBuilder.__init__() missing 1 required positional argument: 'manager'    ```

An interim solution would be to update the requirement:

`'astroid~=3.0;python_version>="3.12"',` Which would allow it to pickup any 3.x build but stop it grabbing the next major one.

Thanks

PR #537, opened 2025-10-05
Update astroid dependency to avoid breaking release

bn-andrew:
Potential fix for #536 
Don't let it grab a breaking version of astroid

Unsure why the version for python3.9-3.11 was so low. Tried changing it to allow any 3.x release


AWhetter:
Declining in favour of #538, which includes support for astroid v4.

PR #541, opened 2025-10-21
Adding option to follow symlinks when generating API

leakec:
* Adding `autoapi_follow_symlinks`, which can optionally be enabled to follow symlinked directories when `os.walk`ing to generate the API docs.
* Added documentation for `autoapi_follow_symlinks`.
* Added unit tests for `autoapi_follow_symlinks`.
* Fixed `mypy` error in `_mapper.py`. This was broken before I made any changes, but fixed it so all the unit tests are passing. I just handled the `None` case by explicitly throwing an error if the `match` returns `None`. This would cause an error in the next line anyway when `match.groups()` was called, since `None` doens't contain `groups()`.

### DIFF
--- a/docs/changes/+7eaa921c.misc.rst
+++ b/docs/changes/+7eaa921c.misc.rst
@@ -1,0 +1,1 @@
+Handling case where match returns None to fix mypy unit test.


AWhetter:
This is perfect! Thank you for the contribution, and for waiting so long for me to review this.

PR #559, opened 2026-05-01
Document the autoapisummary directive

### LINKED ISSUE #377: Document `autoapisummary`
I'm trying to move a docs project from `sphinx.autosummary` to `sphinx-autoapi`, so I'm trying to understand whether I can translate the `.. autosummary` directives with `.. autoapisummary`. However, the latter is not documented. It's defined here:

https://github.com/readthedocs/sphinx-autoapi/blob/85739d513e2dceba4e9c9190aead68b7d9f96870/autoapi/directives.py#L12-L13

Herrtian:
Fixes #377.

This adds a short reference entry for `autoapisummary`, including what it renders, how entries are specified, and a small usage example. I also added a towncrier fragment for the docs change.

Checked with:
- `uv run --with tox tox -e doc,release_notes`
- `git diff --check`

### DIFF
--- a/docs/changes/377.doc.rst
+++ b/docs/changes/377.doc.rst
@@ -1,0 +1,1 @@
+Document the ``autoapisummary`` directive.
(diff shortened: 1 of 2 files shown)


AWhetter:
Thanks for taking the time to make this change!

PR #560, opened 2026-05-05
Document how to render __init__ separately

### LINKED ISSUE #532: Allow separate documentation for class and `__init__` constructor
Context [here](https://github.com/readthedocs/sphinx-autoapi/issues/476#issuecomment-3070409044).

It would be nice to be able to separately document a class and its `__init__` constructor.

This could potentially be done by adding an extra option to [`autoapi_python_class_content`](https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#confval-autoapi_python_class_content), say `separate`.

Herrtian:
Fixes #532.

Adds a short note to `autoapi_python_class_content` explaining that AutoAPI skips `__init__` as a separately rendered method by default, and shows how to use the `autoapi-skip-member` event to include it when the class page should keep only the class docstring.

Checks:
- git diff --check
- uvx --with tox-uv tox -e doc

### DIFF
--- a/doc/source/reference/config.rst
+++ b/doc/source/reference/config.rst
@@ -134,6 +134,26 @@ Customisation Options
    docstring is empty and the class defines a ``__new__`` with a docstring,
    the ``__new__`` docstring is used instead of the ``__init__`` docstring.
 
+   .. note::
+
+      AutoAPI skips ``__init__`` as a separately rendered method by default,
+      even when ``special-members`` is enabled in
+      :confval:`autoapi_options`.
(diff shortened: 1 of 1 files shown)


Herrtian:
Looks like the current macOS 3.12 failure happens before tests while `uv` is installing the built sdist:

```text
Failed to deserialize cache entry
invalid ID: tT9DP-MRCDx1GA1IC4t6T
```

I can't rerun it from here because GitHub requires repo admin rights for that run. The local docs build passed with `uvx --with tox-uv tox -e doc`.

PR #568, opened 2026-08-08
Fix ellipsis rendering in type annotations

### LINKED ISSUE #562: `Concatenate[T, ...]` becomes `Concatenate[T, Ellipsis]`
AutoAPI translates `Callable[Concatenate[T, ...], R]`, valid as per the [typing spec](https://docs.python.org/3/library/typing.html#typing.Concatenate) and meant to represent a function accepting arbitrary parameters, the first of which must be of type `T` for some type expression `T`, and returning a value of type `R`, to `Callable[Concatenate[T, Ellipsis], R]`. This issue propagates when running `sphinx-build` in nitpicky mode, where a `[ref.class]` warning is emitted, signifying `Ellipsis` as a reference target is not found.

Guflly:
Fixes #562. Preserves the ellipsis literal when formatting Concatenate annotations and adds unit and integration coverage.


AWhetter:
Thanks for making this change!

